PR #1, opened 2023-08-05
Start collection of basic data types

djc:
The idea here is to share a long-term stable (as in, no semver-breaking updates) collection of types that can be used as shared context between rustls, webpki, webpki-roots, rustls-native-certs and rustls-pemfile. Right now, there are a number of pain points where the conversion from rustls-native-certs to rustls `OwnedTrustAnchor` is pretty manual. 

Some of these crates (like rustls-native-certs) previously had a dependency on rustls to share its basic `Certificate` and `PrivateKey` dependencies, however rustls evolves much more quickly than rustls-native-certs so that also generated some unwanted churn.

The API here is optimized for long-term stability and doesn't need `std` so it can be used with webpki in `no_std` mode. Its dependency on `alloc` is optional. We also use an inner `Cow`-like type to abstract over slices and `Vec`s with a lifetime, so that a `TrustAnchor` using `&'static [u8]` field values could be used directly in a rustls verifier.

There has been some discussion of this in rustls issues/PRs but I can't find it now. At the time, Brian was not a fan of this idea.

Related PRs:

* https://github.com/rustls/pemfile/pull/24
* https://github.com/rustls/webpki/pull/147
* https://github.com/rustls/webpki-roots/pull/45 (depends on webpki)
* https://github.com/rustls/rustls/pull/1432 (depends on webpki and webpki-roots)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,14 +1,351 @@
-pub fn add(left: usize, right: usize) -> usize {
-    left + right
+//! This crate provides types for representing X.509 certificates, keys and other types as
+//! commonly used in the rustls ecosystem. It is intended to be used by crates that need to work
+//! with such X.509 types, such as [rustls](https://crates.io/crates/rustls),
+//! [rustls-webpki](https://crates.io/crates/rustls-webpki),
+//! [rustls-pemfile](https://crates.io/crates/rustls-pemfile), and others.
+//!
+//! Some of these crates used to define their own trivial wrappers around DER-encoded bytes.
+//! However, in order to avoid inconvenient dependency edges, these were all disconnected. By
+//! using a common low-level crate of types with long-term stable API, we hope to avoid the
+//! downsides of unnecessary dependency edges while providing good interoperability between crates.
+//!
+//! ## DER and PEM
+//!
+//! Many of the types defined in this crate represent DER-encoded data. DER is a binary encoding of
+//! the ASN.1 format commonly used in web PKI specifications. It is a binary encoding, so it is
+//! relatively compact when stored in memory. However, as a binary format, it is not very easy to
+//! work with for humans and in contexts where binary data is inconvenient. For this reason,
+//! many tools and protocols use a ASCII-based encoding of DER, called PEM. In addition to the
+//! base64-encoded DER, PEM objects are delimited by header and footer lines which indicate the type
+//! of object contained in the PEM blob.
+//!
+//! The [rustls-pemfile](https://docs.rs/rustls-pemfile) crate can be used to parse PEM files.
+//!
+//! ## Creating new certificates and keys
+//!
+//! This crate does not provide any functionality for creating new certificates or keys. However,
+//! the [rcgen](https://docs.rs/rcgen) crate can be used to create new certificates and keys.
+
+#![no_std]
+#![warn(unreachable_pub, clippy::use_self)]
+#![deny(missing_docs)]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
+
+#[cfg(feature = "alloc")]
+extern crate alloc;
+
+#[cfg(feature = "alloc")]
+use alloc::vec::Vec;
+use core::fmt;
+use core::ops::Deref;
+
+/// A DER-encoded X.509 private key, in one of several formats
+///
+/// See variant inner types for more detailed information.
+#[non_exhaustive]
+#[derive(Debug, PartialEq)]
+pub enum PrivateKeyDer<'a> {
+    /// An RSA private key
+    Pkcs1(PrivatePkcs1KeyDer<'a>),
+    /// A Sec1 private key
+    Sec1(PrivateSec1KeyDer<'a>),
+    /// A PKCS#8 private key
+    Pkcs8(PrivatePkcs8KeyDer<'a>),
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn it_works() {
-        let result = add(2, 2);
-        assert_eq!(result, 4);
+impl<'a> PrivateKeyDer<'a> {
+    /// Yield the DER-encoded bytes of the private key
+    pub fn secret_der(&self) -> &[u8] {
+        match self {
+            PrivateKeyDer::Pkcs1(key) => key.secret_pkcs1_der(),
+            PrivateKeyDer::Sec1(key) => key.secret_sec1_der(),
+            PrivateKeyDer::Pkcs8(key) => key.secret_pkcs8_der(),
+        }
     }
+}
+
+impl<'a> From<PrivatePkcs1KeyDer<'a>> for PrivateKeyDer<'a> {
+    fn from(key: PrivatePkcs1KeyDer<'a>) -> Self {
+        Self::Pkcs1(key)
+    }
+}
+
+impl<'a> From<PrivateSec1KeyDer<'a>> for PrivateKeyDer<'a> {
+    fn from(key: PrivateSec1KeyDer<'a>) -> Self {
+        Self::Sec1(key)
+    }
+}
+
+impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
+    fn from(key: PrivatePkcs8KeyDer<'a>) -> Self {
+        Self::Pkcs8(key)
+    }
+}
+
+/// A DER-encoded plaintext RSA private key; as specified in PKCS#1/RFC 3447
+///
+/// RSA private keys are identified in PEM context as `RSA PRIVATE KEY` and when stored in a
+/// file usually use a `.pem` or `.key` extension. For more on PEM files, refer to the crate
+/// documentation.
+#[derive(PartialEq)]
+pub struct PrivatePkcs1KeyDer<'a>(Der<'a>);
+
+impl PrivatePkcs1KeyDer<'_> {
+    /// Yield the DER-encoded bytes of the private key
+    pub fn secret_pkcs1_der(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+impl<'a> From<&'a [u8]> for PrivatePkcs1KeyDer<'a> {
+    fn from(slice: &'a [u8]) -> Self {
+        Self(Der(DerInner::Borrowed(slice)))
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl<'a> From<Vec<u8>> for PrivatePkcs1KeyDer<'a> {
+    fn from(vec: Vec<u8>) -> Self {
+        Self(Der(DerInner::Owned(vec)))
+    }
+}
+
+impl fmt::Debug for PrivatePkcs1KeyDer<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("PrivatePkcs1KeyDer")
+            .field(&"[secret key elided]")
+            .finish()
+    }
+}
+
+/// A Sec1-encoded plaintext private key; as specified in RFC 5915
+///
+/// Sec1 private keys are identified in PEM context as `EC PRIVATE KEY` and when stored in a
+/// file usually use a `.pem` or `.key` extension. For more on PEM files, refer to the crate
+/// documentation.
+#[derive(PartialEq)]
+pub struct PrivateSec1KeyDer<'a>(Der<'a>);
+
+impl PrivateSec1KeyDer<'_> {
+    /// Yield the DER-encoded bytes of the private key
+    pub fn secret_sec1_der(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+impl<'a> From<&'a [u8]> for PrivateSec1KeyDer<'a> {
+    fn from(slice: &'a [u8]) -> Self {
+        Self(Der(DerInner::Borrowed(slice)))
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl<'a> From<Vec<u8>> for PrivateSec1KeyDer<'a> {
+    fn from(vec: Vec<u8>) -> Self {
+        Self(Der(DerInner::Owned(vec)))
+    }
+}
+
+impl fmt::Debug for PrivateSec1KeyDer<'_> {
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -1,14 +1,259 @@
-pub fn add(left: usize, right: usize) -> usize {
-    left + right
+#![no_std]
```
cpu, line 1: Does using `no_std` like this mean we don't have to include the `clippy::alloc_instead_of_core` lint from https://github.com/rustls/rustls/pull/1378?

src/lib.rs
```diff
@@ -1,14 +1,259 @@
...
+#![warn(unreachable_pub)]
```
cpu, line 2: Did you consider matching the linting config from the main rustls repo?

Over there we treat `deny` `unreachable_pub` and also include some others like `missing_docs`, `clippy::use_self` etc. I think some of those might be worthwhile here too.

cpu (commented):
Is the idea here that `rustls`, `webpki`, `webpki-roots` and `rustls-pemfile` will all share a dependency on this crate for interoperable types? Could you update the PR description with a bit of the higher-level rationale? (I think some of it was discussed in another issue/PR but I can't find it off-hand)

I left a couple comments about clippy settings, but otherwise the code in here seems reasonable for my understanding of what we're trying to achieve with the separate repo.

src/lib.rs
djc: I'm not super familiar with `no_std` setups, but given that so far in this crate nothing really needs `std`, it seemed most straightforward to unconditionally opt-in to `no_std` and make the `alloc` dependency conditional. Given the CI setup I think the lint is superfluous, because `--no-default-features` should trigger the same failures.

src/lib.rs
djc: I kind of wanted to start from scratch because there are some downsides to cargo-culting around the setup from older repos. A bunch of the lints that we are setup to `allow` in rustls seem like they are no longer enabled by default, for example, and all other things being equal I'd rather have smaller lint deviants from default.

In general for most of these things I prefer avoiding `deny` and using `warn` instead while denying warnings in CI only.

So this is more like my "native" minimal lint setup. Extending this with more stuff could make sense but is low on my priority list relative to finding a good API that can serve all the crates.

djc:
Would also like feedback on this from @jsha (concerning FFIability) and @complexspaces (regarding usability within rustls-platform-verifier).

src/lib.rs
cpu: That makes sense to me, thanks for explaining :+1:

ctz:
I think this is looking good so far -- my main comment is that we should be making clear somewhere that these are relatively behaviour-less newtypes, so nobody should end up with an expectation that (for example) `CertificateDer` contains any particular encoding.

jsha:
For convenient reference I've pushed docs from this PR to https://rustdoc.crud.net/jsha/pki-types/rustls_pki_types/.

One thing I'm wondering: are these newtypes even necessary? Can we pass primitive slices like `&[u8]` in their place? In general newtypes are nice for preventing mistakes (like passing a key instead of a certificate). But I think for these people just convert bytes to the newtype and gain no real safety. For instance, look at the [example for Certificate](https://docs.rs/rustls/latest/rustls/struct.Certificate.html):

```
fn load_certificates_from_pem(path: &str) -> std::io::Result<Vec<Certificate>> {
    let file = File::open(path)?;
    let mut reader = BufReader::new(file);
    let certs = rustls_pemfile::certs(&mut reader)?;

    Ok(certs.into_iter().map(Certificate).collect())
}
```

The `map(Certificate)` here doesn't seem to add very much safety or clarity. Though I suppose the idea is that `rustls_pemfile::certs` would return `Vec<pki_types::CertificateDer<'a>>` instead of `Vec<Vec<u8>>` as it does now? 

Also to make sure I understand correctly: The Cow-alike implementation here (DerInner::Slice and DerInner::Vec) is not a performance optimization, it's to allow operation in a `no_std` environment that also does not have the `alloc` library, because `Vec` is not available in that environment? And in that environment only the `DerInner::Slice` variant would be available? As a concrete example: on an embedded system, someone might know that some specific 4kB region of memory is available, make a slice out of it (unsafely but with caution), read some bytes into it, and then make a `CertificateDer` that references that slice?

Assuming all the above is correct, I feel like I've talked myself into understanding why these are necessary, but I'll still include the ruminating since it may be useful to have it written out (and I could be wrong).

From an FFI standpoint I think this should be fine. We currently expose the `Certificate` type as `rustls_certificate`: https://github.com/rustls/rustls-ffi/blob/db40a7b8cee883b131d41741135e440a98a0b0eb/src/cipher.rs#L26-L62

That's a simple struct that has one accessor, to get a pointer and length to the contained data. It would probably keep the same interface. It's returned from `rustls_connection_get_peer_certificate` and `rustls_certified_key_get_certificate`. These both document the returned pointer's lifetime to be limited by some parent object, so they should be okay with the slice variant.

The other place I would expect to see the `rustls_certificate` type is [rustls_verify_server_cert_params](https://docs.rs/rustls-ffi/latest/rustls_ffi/client/struct.rustls_verify_server_cert_params.html), passed into certificate verification callbacks. But actually the current version just contains a `rustls_slice_bytes`, which seems reasonable; in the callback, there's always a bounding lifetime (that is, the duration of the callback).

rustls-ffi doesn't current expose TrustAnchor at all. Instead it exposes RootCertStore as rustls_root_cert_store, and offers [rustls_root_cert_store_add_pem](https://github.com/rustls/rustls-ffi/blob/db40a7b8cee883b131d41741135e440a98a0b0eb/src/cipher.rs#L465-L500), which calls add_parseable_certificates.

One slight reservation I have on the FFI front is: will there ever be a case when we want non-Rust code to be able to generate the Slice / borrowed variant of `CertificateDer` or others? I worry about that scenario because it requires the non-Rust code to carefully manage lifetimes. But I think if it does become necessary we can always document the lifetime requirements, which are not really any more onerous than what C code deals with all the time. And if we want to eventually support `no_std` + no `alloc` crate + FFI, it would be necessary.

Oh and one thing to be aware of on the FFI front: `&[u8]` is nice and easy because we don't worry about `#[repr(C)]`. We break it down into a pointer and len and give those to the C side as needed. `&[Vec<u8>]` and `&[CertificateDer<'a>]` are challenging because we can't just pass a pointer and len; the representation of `Vec` and `CertificateDer` are not guaranteed. For `rustls_verify_server_cert_callback` in rustls-ffi we handled this in a hacky way by collecting a `&[&[u8]]` in the Rust function before calling the callback. However, I think a more general solution is possible, and preferred: any time we have, e.g. `&[CertificateDer<'a>]`, we need to treat that as its own FFI type. For instance `*const rustls_slice_certificate_der`. And then to get any given element from that slice, you would call an accessor, e.g. `rustls_slice_certificate_der_get_ith(i)`. That would return `*rustls_certificate_der`, which would have its own accessor `rustls_certificate_der_as_slice()` (returning `rustls_slice_bytes`).

So, overall after chewing it over for a bit: 👍🏻

djc:
Thanks for the thorough review!

> The `map(Certificate)` here doesn't seem to add very much safety or clarity. Though I suppose the idea is that `rustls_pemfile::certs` would return `Vec<pki_types::CertificateDer<'a>>` instead of `Vec<Vec<u8>>` as it does now?

Yes, it would return a `Vec<CertificateDer<'static>>` -- the same would go for `rustls_native_certs::load_native_certs()`.

> Also to make sure I understand correctly: The Cow-alike implementation here (DerInner::Slice and DerInner::Vec) is not a performance optimization, it's to allow operation in a `no_std` environment that also does not have the `alloc` library, because `Vec` is not available in that environment? And in that environment only the `DerInner::Slice` variant would be available? As a concrete example: on an embedded system, someone might know that some specific 4kB region of memory is available, make a slice out of it (unsafely but with caution), read some bytes into it, and then make a `CertificateDer` that references that slice?

I think there are two goals:

* Make it possible to make these data types work in a no-`alloc` environment
* Abstract over the use of the data type (that is, the difference between `Vec<u8>` and `&'static [u8]`

For the latter, I'm thinking of scenarios where a `TrustAnchor` or `Certificate` might be embedded in a library or application as a `const` (as in webpki-roots), and allocating for it is actually unnecessary.

> rustls-ffi doesn't current expose TrustAnchor at all. Instead it exposes RootCertStore as rustls_root_cert_store, and offers [rustls_root_cert_store_add_pem](https://github.com/rustls/rustls-ffi/blob/db40a7b8cee883b131d41741135e440a98a0b0eb/src/cipher.rs#L465-L500), which calls add_parseable_certificates.

So one potential future direction that I think might be interesting is removing the `RootCertStore` abstraction in favor of just `Vec<TrustAnchor<'static>>` -- because I think the `RootCertStore` abstraction isn't providing a whole lot of value.

jsha:
> > Abstract over the use of the data type (that is, the difference between Vec<u8> and &'static [u8]
>
> For the latter, I'm thinking of scenarios where a TrustAnchor or Certificate might be embedded in a library or application as a const (as in webpki-roots), and allocating for it is actually unnecessary.

Ah, nice. I think the case of embedding a `CertificateDer` in an application would be quite unusual for a couple of reasons. It means rebuilding and redeploying is required to rotate certificates. Also, the `CertificateDer` is not much good without a corresponding private key, and it would be bad practice to embed a private key in an application.

The TrustAnchor case is of course sensible and we have a current example of embedding a list of TrustAnchors in a library (webpki-roots). I increasingly worry about the reliability hazard posed by applications that bake in a list of TrustAnchors and don't get updated, but that's not really an issue for this thread (and hopefully we'll solve it by promoting rustls-platform-verifier as the best way).

djc:
> I increasingly worry about the reliability hazard posed by applications that bake in a list of TrustAnchors and don't get updated, but that's not really an issue for this thread (and hopefully we'll solve it by promoting rustls-platform-verifier as the best way).

Yup, I'm still on board with promoting rustls-platform-verifier as the better verifier, but I also think there are use cases where webpki-roots are quite nice. In general, I would argue rustls-platform-verifier is especially useful on Windows and macOS as I don't have as (a) on Linux we fall back to the rustls-native-certs approach which (b) depends on distribution trust store management, which might well be less reliable than the Mozilla roots. Obviously that still depends on timely upgrades of webpki-roots and regular build + deployments, but I think that scenario still covers a lot of ground in SaaS applications.

djc:
Update:

* Added `Debug` implementations for the private key types (which do not print the private key bytes)
* Derived `PartialEq` for all the types

Of course the `PartialEq` implementation for private key types effectively also leaks the private key, but in a less bad way than accidentally leaking it through a derived `Debug` impl of a downstream type containing it. The `PartialEq` requirement comes from `rustls_pemfile::Item` -- it was added in the [Initial commit](https://github.com/rustls/pemfile/commit/e5b5970d7baf150af99f98ab99dd0de06e840438) so not sure how load-bearing that is.

src/lib.rs
```diff
@@ -1,14 +1,331 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("Der").field(&self.as_ref()).finish()
     }
 }
+
+#[derive(Clone, PartialEq)]
+enum DerInner<'a> {
+    #[cfg(feature = "alloc")]
+    Vec(Vec<u8>),
+    Slice(&'a [u8]),
```
ctz, line 330: I would have called these variants `Owned` and `Borrowed` but both work for me.

ctz (commented):
This is looking good to me.

One of the things I took from the rustls no-std RFC was whether a unix timestamp newtype would be good, to hoist up that type from webpki into here. Do you think that's useful? I suppose don't need to make that decision now in any case.

djc:
> One of the things I took from the rustls no-std RFC was whether a unix timestamp newtype would be good, to hoist up that type from webpki into here. Do you think that's useful? I suppose don't need to make that decision now in any case.

Yes, that's definitely something that I think would make sense, but trying to deal with the initial push here first, which is complicated enough when working across 6 different repos.

src/lib.rs
djc: Makes sense, I can change them.

cpu (approved):
This looks good to me. 

Would you be willing to take some of the context you added to the PR description and using it to create a `README.md`? I think it would be useful to capture the purpose of the crate there.

Otherwise my only feedback was related to comments you were porting as-is from existing code. If you don't want to adjust those as part of this work that's fine with me. Re-reading the comments in this branch prompted some thoughts for potential updates but I don't think they need to be blocking.

src/lib.rs
```diff
@@ -1,14 +1,331 @@
...
+/// like this:
+///
+/// ```txt
+/// -----BEGIN BEGIN EC PRIVATE KEY-----
+/// <base64-encoded certificate content>
+/// -----END CERTIFICATE-----
+/// ```
+///
+/// The [`rustls-pemfile`](https://docs.rs/rustls-pemfile/latest/rustls_pemfile/) crate can be used
+/// to parse PEM files. The [`rcgen`](https://docs.rs/rcgen/latest/rcgen/) crate can be used to
```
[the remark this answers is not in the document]
djc: I've removed some of the repeated PEM contents from the types' documentation in favor of a crate root explanation.

src/lib.rs
```diff
@@ -1,14 +1,331 @@
...
+impl fmt::Debug for PrivatePkcs8KeyDer<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("PrivatePkcs1KeyDer")
+            .field(&"[secret key elided]")
+            .finish()
+    }
+}
+
+/// A trust anchor (a.k.a. root CA), represented as a DER-encoded X.509 certificate
+///
```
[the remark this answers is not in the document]
djc: Agreed.

djc:
Addressed a bunch of documentation issues, added a README and crate root documentation, added `#![deny(missing_docs)]` to make sure all public API is documented and `#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]` to make sure the `alloc`-based API are documented as such. I think this is ready to publish as a 0.1 release.

ctz (approved):

src/lib.rs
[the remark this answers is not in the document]
cpu: Thanks, looks good.

PR #4, opened 2023-09-04
Add TLSA types that will be used to implement DANE

wez:
I put these in a separate module to avoid clutter, but that can be moved if you prefer to use a different style for this.

One thing I'm not totally sure about is this paragraph at the bottom of https://datatracker.ietf.org/doc/html/rfc6698#section-2.1.1:

> The certificate usages defined in this document explicitly only apply
   to PKIX-formatted certificates in DER encoding [[X.690](https://datatracker.ietf.org/doc/html/rfc6698#ref-X.690)].  If TLS
   allows other formats later, or if extensions to this RRtype are made
   that accept other formats for certificates, those certificates will
   need their own certificate usage values.

For the sake of simplicity in constructing the `TlsaRecord` I've encoded the data field as `Der`, but the above suggests that perhaps the `CertificateUsage` variants should become "load bearing" members that have `Der` data inside them.

Thoughts?

refs: https://github.com/rustls/rustls/issues/816

### DIFF
--- a/src/tlsa.rs
+++ b/src/tlsa.rs
@@ -1,0 +1,120 @@
+//! This types in this module relate to the TLSA and DANE
+//! as described by [RFC6698](https://datatracker.ietf.org/doc/html/rfc6698)
+use crate::Der;
+
+/// Corresponds to the TLSA DNS RR Type as defined by
+/// [RFC6698](https://datatracker.ietf.org/doc/html/rfc6698)
+#[derive(Clone, Debug, PartialEq)]
+pub struct TlsaRecord<'a> {
+    /// Specifies the provided association that will be used to match
+    /// the certificate presented in the TLS handshake.
+    pub certificate_usage: CertificateUsage,
+
+    /// specifies which part of the TLS certificate presented by the
+    /// server will be matched against the association data
+    pub selector: Selector,
+
+    /// Specifies how the certificate association is presented.
+    pub matching: MatchingType,
+
+    /// This field specifies the "certificate association data" to be
+    /// matched.  These bytes are either raw data (that is, the full
+    /// certificate or its SubjectPublicKeyInfo, depending on the selector)
+    /// for matching type 0, or the hash of the raw data for matching types 1
+    /// and 2.  The data refers to the certificate in the association, not to
+    /// the TLS ASN.1 Certificate object.
+    pub association_data: Der<'a>,
+}
+
+/// A TLSA Certificate Usage as described in [RFC6698 section
+/// 7.2](https://datatracker.ietf.org/doc/html/rfc6698#section-7.2).
+/// This may be extended by future RFCs and is thus marked as
+/// non-exhaustive.
+#[repr(u8)]
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum CertificateUsage {
+    /// Certificate usage 0 is used to specify a CA certificate, or
+    /// the public key of such a certificate, that MUST be found in any of
+    /// the PKIX certification paths for the end entity certificate given
+    /// by the server in TLS.  This certificate usage is sometimes
+    /// referred to as "CA constraint" because it limits which CA can be
+    /// used to issue certificates for a given service on a host.  The
+    /// presented certificate MUST pass PKIX certification path
+    /// validation, and a CA certificate that matches the TLSA record MUST
+    /// be included as part of a valid certification path.  Because this
+    /// certificate usage allows both trust anchors and CA certificates,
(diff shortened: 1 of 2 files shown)


src/tlsa.rs
```diff
@@ -0,0 +1,120 @@
+//! This types in this module relate to the TLSA and DANE
```
wez, line 1: some typos in these docs; will make a pass to fix these up

wez:
A couple of problems with this, if `TlsaRecord` is to be embedded in `rustls::ServerName`:

* `ServerName` is `Hash+Eq` but `Der` is not. I can make Der `Hash+Eq`, seemingly without consequence. I'm not sure that it is valid to relax `ServerName` to be just `PartialEq`; I'm not sure how/where `ServerName` is hashed.
* The lifetime in `Der<'a>` either needs to infect `ServerName<'a>` or requires the use of `'static` when embedded in there.

src/tlsa.rs
```diff
@@ -0,0 +1,120 @@
...
+    CAConstraint = 0,
+    /// Certificate usage 1 is used to specify an end entity
+    /// certificate, or the public key of such a certificate, that MUST be
+    /// matched with the end entity certificate given by the server in
+    /// TLS.  This certificate usage is sometimes referred to as "service
+    /// certificate constraint" because it limits which end entity
+    /// certificate can be used by a given service on a host.  The target
+    /// certificate MUST pass PKIX certification path validation and MUST
+    /// match the TLSA record.
+    ServiceCertificateConstraint = 1,
```
djc, line 58: To the extent that this documentation is quoting directly from the spec, perhaps make that clear by using quotes of some kind and linking to the originating section. It feels like there is a lot of boilerplate/repetition here that is making it harder to actually understand the core of the difference between these variants?

src/tlsa.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// Specifies how the certificate association is presented.
+/// The value is defined in an IANA registry is and described in [RFC6698 Section
+/// 7.4](https://datatracker.ietf.org/doc/html/rfc6698#section-7.4)
+#[repr(u8)]
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MatchingType {
+    /// Exact match on selected content
+    ExactMatch = 0,
```
djc, line 113: Nit: I'd probably avoid repeating the `Match` suffix from the type name.

src/tlsa.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Specifies how the certificate association is presented.
+    pub matching: MatchingType,
+
+    /// This field specifies the "certificate association data" to be
+    /// matched.  These bytes are either raw data (that is, the full
+    /// certificate or its SubjectPublicKeyInfo, depending on the selector)
+    /// for matching type 0, or the hash of the raw data for matching types 1
+    /// and 2.  The data refers to the certificate in the association, not to
+    /// the TLS ASN.1 Certificate object.
+    pub association_data: Der<'a>,
```
djc, line 26: As you alluded to, I don't think this is always `Der<'_>`, at least I assume that in the hash cases these would just be straight up bytes?

How do you feel about modeling it more like this?

```rust
enum CertificateUsage {
    CaConstraint(Selector),
    ..
}

enum Selector {
    FullCertificate(MatchingType),
}

enum MatchingType {
     Exact(Der<'_>),
     Sha256([u8; 32]),
     Sha512([u8; 64]),
}
```

djc:
> A couple of problems with this, if `TlsaRecord` is to be embedded in `rustls::ServerName`:
> 
>     * `ServerName` is `Hash+Eq` but `Der` is not. I can make Der `Hash+Eq`, seemingly without consequence. I'm not sure that it is valid to relax `ServerName` to be just `PartialEq`; I'm not sure how/where `ServerName` is hashed.

Try removing the `Hash` bound and see what fails? I think it's probably okay to make `Der` `Hash`, too.

>     * The lifetime in `Der<'a>` either needs to infect `ServerName<'a>` or requires the use of `'static` when embedded in there.

I think we'll want to let it infect `ServerName`? IIRC `ServerName` so far is a rustls concept, might be worth seeing how we represent this in rustls-webpki (where, IIRC, we avoid allocating on the common paths).

cpu:
The path forward here feels uncertain enough that it might make sense to make this a draft PR and consider it for merging only after having used it in downstream draft pull requests that specify this branch as a [dependency override](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html). I suspect there pursuing the webpki and rustls updates will affect the shape of the common types.

PR #5, opened 2023-09-06
Migrate signature verification trait to here

ctz:
The goal of this is that downstream crates can implement and pass around this trait and associated types without taking a dependency on a webpki version.

This seems necessary to avoid changes like https://github.com/rustls/rustls/pull/1405/commits/7462764643e62acc763eb51a6bbbf305d5cfa51c exposing webpki types in rustls's public API.

This is not a direct lift-and-shift, the types have changed for consistency with others here, eg `AlgorithmIdentifier` reuses `Der`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -291,6 +291,72 @@ impl<'a> From<Vec<u8>> for CertificateDer<'a> {
     }
 }
 
+/// A DER encoding of the PKIX AlgorithmIdentifier type:
+///
+/// ```ASN.1
+/// AlgorithmIdentifier  ::=  SEQUENCE  {
+///     algorithm               OBJECT IDENTIFIER,
+///     parameters              ANY DEFINED BY algorithm OPTIONAL  }
+///                                -- contains a value of the type
+///                                -- registered for use with the
+///                                -- algorithm object identifier value
+/// ```
+/// (from <https://www.rfc-editor.org/rfc/rfc5280#section-4.1.1.2>)
+///
+/// The outer sequence encoding is *not included*, so this is the DER encoding
+/// of an OID for `algorithm` plus the `parameters` value.
+///
+/// For example, this is the `rsaEncryption` algorithm:
+///
+/// ```
+/// let rsa_encryption = rustls_pki_types::AlgorithmIdentifier::from_slice(
+///     &[
+///         // algorithm: 1.2.840.113549.1.1.1
+///         0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x01,
+///         // parameters: NULL
+///         0x05, 0x00
+///     ]
+/// );
+/// ```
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -349,3 +349,69 @@ enum DerInner<'a> {
...
+///     &[
+///         // algorithm: 1.2.840.113549.1.1.1
+///         0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x01,
+///         // parameters: NULL
+///         0x05, 0x00
+///     ]
+/// );
+/// ```
+#[derive(Debug, Clone, PartialEq)]
+pub struct AlgorithmIdentifier<'a>(Der<'a>);
```
djc, line 426: Should we rename this to `SignatureAlgorithmIdentifier` instead?

src/lib.rs
```diff
@@ -350,6 +350,51 @@ enum DerInner<'a> {
...
 
+/// An abstract signature verification algorithm.
+///
+/// One of these is needed per supported pair of public key type (identified
+/// with `public_key_alg_id()`) and `signatureAlgorithm` (identified with
+/// `signature_alg_id()`).  Note that both of these `AlgorithmIdentifier`s include
+/// the parameters encoding, so separate `SignatureVerificationAlgorithm`s are needed
+/// for each possible public key or signature parameters.
+pub trait SignatureVerificationAlgorithm: Send + Sync {
+    /// Return the `AlgorithmIdentifier` that must be present on a `subjectPublicKeyInfo`
```
djc, line 361: Nit: I'd invert the order on these methods.

djc (approved):
LGTM!

djc:
(I'd also move both of these up before the `Der` definition.)

src/lib.rs
cpu: Since it parameters in addition to OID I think that makes sense.

cpu (approved):

src/lib.rs
ctz: i'm minded to keep `AlgorithmIdentifier` to match the naming in RFC5280. plus -- annoyingly -- there's not a separate namespace of these used in signatures and public keys.

src/lib.rs
```diff
@@ -349,3 +349,114 @@ enum DerInner<'a> {
...
+    /// for this `SignatureVerificationAlgorithm` to be considered for verification.
+    fn public_key_alg_id(&self) -> AlgorithmIdentifier;
+
+    /// Return the `AlgorithmIdentifier` that must be present as the `signatureAlgorithm`
+    /// on the data to be verified for this `SignatureVerificationAlgorithm` to be considered
+    /// for this `SignatureVerificationAlgorithm` to be considered.
```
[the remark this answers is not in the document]
ctz: good spot, have reworded this and the other comment here

djc (approved):

djc:
(You'll probably want a version bump?)

PR #7, opened 2023-09-12
Add UnixTime type

djc:
As proposed in https://github.com/rustls/rustls/pull/1399#discussion_r1293648324.

I propose we replace `webpki::Time` and `rustls::ticketer::TimeBase` with this type. I also propose we replace the use of `SystemTime` in rustls' `Verifier` interfaces with this type. Code that depend on `std` (including current mainline rustls) can trivially call `UnixTime::now()` instead of `SystemTime::now()`.

r? @japaric


ctz (approved):
👍🏽

src/lib.rs
```diff
@@ -391,6 +395,32 @@ impl Deref for AlgorithmIdentifier {
     }
 }
 
+/// A timestamp, tracking the number of non-leap seconds since the start of 1970
```
japaric, line 398: I think it would be good to mention the term "Unix epoch" in the API docs as that's well understood to be `1970-01-01T00:00:00+00:00` (timezone matters)

japaric (approved):
thanks! this looks good to me. left an API doc suggestion

djc:
Yes, I struggled a bit with naming. In principle, I liked the notion of avoiding the concept of the Unix epoch and expressing everything in terms of a normal calendar date epoch, but maybe that's just too verbose and we should place the Unix heritance front and center (which I somewhat did already by naming it `UnixTime`). So should we also rename `since_1970()` to `since_unix_epoch()`?

cpu (approved):

djc:
Published 0.2.1.

PR #8, opened 2023-11-14
derive clone for private key

zh-jq:


zh-jq:
How about add a to_owned() method? This can work for us as well.

djc:
I propose we do `into_owned()`, which consumes `self`.

zh-jq:
I solved this by using a [owned struct](https://github.com/bytedance/g3/pull/132/files#diff-7c01a78586826d51d15a266405d760653c76706dc929f958c946d035d4786663R23) in our own code. So close this.

PR #11, opened 2023-11-21
Server name

djc:


src/server_name.rs
```diff
@@ -257,6 +253,277 @@ enum State {
     Hyphen { len: usize },
 }
 
+/// `no_std` implementation of `std::net::IpAddr`.
```
ctz, line 256: Once `std::net::IpAddr` and friends are moved to `core::net::IpAddr` I am hoping this can be lost.

ctz (approved):

src/server_name.rs
djc: Yup, that's the idea.

cpu (approved):

src/server_name.rs
```diff
@@ -0,0 +1,379 @@
...
+}
+
+enum State {
+    Start,
+    Next,
+    NumericOnly { len: usize },
+    NextAfterNumericOnly,
+    Subsequent { len: usize },
+    Hyphen { len: usize },
+}
```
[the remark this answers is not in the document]
djc: Yeah -- I'm not a big fan of inlining such helper types because they distract from the actual logic IMO. But have done so here anyway.

src/server_name.rs
```diff
@@ -0,0 +1,379 @@
...
+    fn try_from(value: &'a [u8]) -> Result<Self, Self::Error> {
+        match str::from_utf8(value) {
+            Ok(s) => Self::try_from(s),
+            Err(_) => return Err(InvalidDnsNameError),
+        }
+    }
+}
+
+/// Attempt to make a ServerName from a string by parsing
+/// it as a DNS name.
```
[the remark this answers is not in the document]
djc: Adapted.

src/server_name.rs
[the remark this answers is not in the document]
cpu: Ok! I don't feel super strongly so in the future knowing you prefer this style I'll avoid similar feedback.

src/server_name.rs
jsha: Perhaps the semver bump could be avoided by making the `IpAddr` value of `ServerName` private? Are there any cases where the application needs to read out that value?

Alternately: semver is about what interfaces are documented to be stable, so we can document our way out of this by saying that the type of the `IpAddr` value is explicitly not stable (yet).

djc: > Perhaps the semver bump could be avoided by making the `IpAddr` value of `ServerName` private? Are there any cases where the application needs to read out that value?

Well, webpki needs to match `ServerName` to get out the DNS name bytes or the IP address bytes, so I don't think hiding this type is straightforward.

> Alternately: semver is about what interfaces are documented to be stable, so we can document our way out of this by saying that the type of the `IpAddr` value is explicitly not stable (yet).

That feels like a reasonable compromise. We can say that the interfaces to build a `ServerName` are stable but that the identity of the `IpAddr` type is not.

djc:
I've added two commits:

* One commit to replace the IP parsing code from webpki with IP parsing code from core. This supports a wider range of input -- as far as I can tell, only for IPv6, where compressed addresses were out of scope for webpki's implementation. Note that rustls was already [relying](https://github.com/rustls/rustls/blob/e9316dfcd49e91ee1d741835cb0c5196151c51ca/rustls/src/client/client_conn.rs#L424) on the core/std implementation for parsing anyway. This also removes a bunch of tests that checked that the implementation indeed failed to parse compressed IPv6 addresses.
* One commit to add notes to the docstrings for `IpAddr`, `Ipv4Addr` and `Ipv6Addr` to exempt their identity from semver-stability, so that we can reserve the right to swap them out for the actual `core::net` types when those are stabilized without requiring a semver-incompatible version bump.

cpu (approved):
The recent updates LGTM

PR #12, opened 2023-11-21
add CertificateDer::into_owned

japaric:
helps with the performance improvements in rustls/rustls#1597 as it's not possible to do this efficiently in the `Owned` case using only the public API of `CertificateDer`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -295,6 +295,14 @@ impl<'a> From<Vec<u8>> for CertificateDer<'a> {
     }
 }
 
+impl CertificateDer<'_> {
+    /// Converts this certificate into its owned variant, unfreezing borrowed content (if any)
+    #[cfg(feature = "alloc")]
+    pub fn into_owned(self) -> CertificateDer<'static> {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -491,3 +499,14 @@ enum DerInner<'a> {
...
 }
+
+#[cfg(feature = "alloc")]
+impl DerInner<'_> {
+    fn into_owned(self) -> DerInner<'static> {
+        let vec = match self {
+            Self::Owned(vec) => vec,
+            Self::Borrowed(slice) => slice.to_vec(),
+        };
+        DerInner::Owned(vec)
```
cpu, line 510: I think this is a little bit cleaner all-in-one:

```suggestion
        DerInner::Owned(match self {
            Self::Owned(vec) => vec,
            Self::Borrowed(slice) => slice.to_vec(),
        })
```

cpu (approved):
Thanks!

japaric:
addressed review comment

djc (approved):

PR #13, opened 2023-11-23
Improve Debug impls

jsha:
For Der and AlgorithmIdentifier, emit hex bytes rather than decimal digits separated by commas. The decimal output is impossible to read and challenging to convert into a better format. For example:

[ ... 105, 103, 105, 67, 101, 114, 116, 44, 32, 73, ... ]

Also, add a Debug bound on SignatureVerificationAlgorithm. This helps upstream implement Debug in a way that can list out which SignatureVerificationAlgorithms are in use.

This copies a technique we added to rustls in https://github.com/rustls/rustls/pull/1040.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -310,7 +310,12 @@ impl CertificateDer<'_> {
 /// `signature_alg_id()`).  Note that both of these `AlgorithmIdentifier`s include
 /// the parameters encoding, so separate `SignatureVerificationAlgorithm`s are needed
 /// for each possible public key or signature parameters.
-pub trait SignatureVerificationAlgorithm: Send + Sync {
+///
+/// Debug implementations should list the public key algorithm identifier and
+/// signature algorithm identifier in human friendly form (i.e. not encoded bytes),
+/// along with the name of the implementing library (to distinguish different
+/// implementations of the same algorithms).
+pub trait SignatureVerificationAlgorithm: Send + Sync + fmt::Debug {
     /// Verify a signature.
     ///
     /// `public_key` is the `subjectPublicKey` value from a `SubjectPublicKeyInfo` encoding
@@ -377,7 +382,7 @@ pub struct InvalidSignature;
 ///     ]
 /// );
 /// ```
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Clone, Copy, PartialEq, Eq)]
 pub struct AlgorithmIdentifier(&'static [u8]);
 
 impl AlgorithmIdentifier {
@@ -392,6 +397,12 @@ impl AlgorithmIdentifier {
 impl AsRef<[u8]> for AlgorithmIdentifier {
     fn as_ref(&self) -> &[u8] {
         self.0
+    }
+}
+
+impl fmt::Debug for AlgorithmIdentifier {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        hex(f, self.0)
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -509,3 +520,26 @@ impl DerInner<'_> {
...
+// Format an iterator of u8 into a hex string
+fn hex<'a>(f: &mut fmt::Formatter<'_>, payload: impl IntoIterator<Item = &'a u8>) -> fmt::Result {
+    for b in payload {
+        write!(f, "{:02x}", b)?;
+    }
+    Ok(())
+}
+
+#[cfg(feature = "std")]
+mod tests {
```
djc, line 533: Nit: I would `use super::*` here so we can avoid the `super::` qualifications inside the tests.

src/lib.rs
```diff
@@ -509,3 +520,26 @@ impl DerInner<'_> {
...
+    #[test]
+    fn der_debug() {
+        let der = super::Der::from_slice(&[0x01, 0x02, 0x03]);
+        assert_eq!(format!("{:?}", der), "010203");
+    }
+
+    #[test]
+    fn algid_debug() {
```
djc, line 541: Nit: I think `alg_id` would be a more idiomatic spelling.

src/lib.rs
```diff
@@ -481,7 +492,7 @@ impl From<Vec<u8>> for Der<'static> {
 
 impl fmt::Debug for Der<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_tuple("Der").field(&self.as_ref()).finish()
+        hex(f, self.as_ref())
```
djc, line 495: Did you consider leaving the `Der(..)` wrapping in? Why/why not?

src/lib.rs
```diff
@@ -509,3 +520,26 @@ impl DerInner<'_> {
         })
     }
 }
+
+// Format an iterator of u8 into a hex string
+fn hex<'a>(f: &mut fmt::Formatter<'_>, payload: impl IntoIterator<Item = &'a u8>) -> fmt::Result {
+    for b in payload {
```
djc, line 526: How do we feel about starting this with a `0x` prefix?

djc (approved):

cpu (approved):

cpu:
I pushed [a commit](3c7f545) addressing most of the review feedback that can be squash merged.

src/lib.rs
cpu: Fixed.

src/lib.rs
cpu: Fixed.

src/lib.rs
cpu: I left this unchanged, both options seem reasonable to me.

src/lib.rs
cpu: Added the prefix.

cpu:
> that can be squash merged.

Ah, this repo only allows the merge queue. I'll squash it manually.

src/lib.rs
jsha: It winds up with multiple layers of wrapping in the Debug output, which gets kind of noisy. For instance, since `CertificateDer` is:

```rust
pub struct CertificateDer<'a>(Der<'a>);
```

It would output as `CertificateDer ( Der ( 03040506 ) )`. To most readers the inner `Der` wrapping is an implementation detail and not that important. But I don't feel strongly about it.

jsha:
Thanks for picking up the review changes, @cpu!

src/lib.rs
djc: Fair enough!

PR #14, opened 2023-11-24
server_name: offer ServerName::to_str w/ alloc/std features

cpu:
Downstream crates may want to represent a `ServerName` as a `String` or `&str`, e.g. for passing through a `ffi` boundary to external code (for example, [in `rustls-ffi`](https://github.com/rustls/rustls-ffi/blob/a26a3d4ad6cbc13670b6ddbe924db075540c92ac/src/client.rs#L263-L267), or in [`rustls-platform-verifier`](https://github.com/rustls/rustls-platform-verifier/blob/e4194174f2b4b235ed13fc057ce89b19894321e6/src/verification/android.rs#L276-L283)).

This commit implements support for converting a `pki_types::IpAddr` to a `std::net::IpAddr` (to use the `to_string` impl), and adds a `ServerName::to_str` that can return a `Cow<'_, str>` for a `DnsName` or an `IpAddr`.

### DIFF
--- a/src/server_name.rs
+++ b/src/server_name.rs
@@ -1,5 +1,7 @@
 //! DNS name validation according to RFC1035, but with underscores allowed.
 
+#[cfg(all(feature = "alloc", feature = "std"))]
+use alloc::borrow::Cow;
 #[cfg(feature = "alloc")]
 use alloc::string::{String, ToString};
 use core::hash::{Hash, Hasher};
@@ -47,6 +49,18 @@ impl<'a> ServerName<'a> {
         match self {
             Self::DnsName(d) => ServerName::DnsName(d.to_owned()),
             Self::IpAddress(i) => ServerName::IpAddress(*i),
+        }
+    }
+
+    /// Return the string representation of this `ServerName`.
+    ///
+    /// In the case of a `ServerName::DnsName` instance, this function returns a borrowed `str`.
+    /// For a `ServerName::IpAddress` instance it returns an allocated `String`.
+    #[cfg(feature = "std")]
+    pub fn to_str(&self) -> Cow<'_, str> {
+        match self {
+            Self::DnsName(d) => d.as_ref().into(),
+            Self::IpAddress(i) => std::net::IpAddr::from(*i).to_string().into(),
         }
     }
 }
@@ -313,6 +327,16 @@ impl From<std::net::IpAddr> for IpAddr {
     }
 }
 
+#[cfg(feature = "std")]
+impl From<IpAddr> for std::net::IpAddr {
+    fn from(value: IpAddr) -> Self {
+        match value {
+            IpAddr::V4(v4) => Self::from(std::net::Ipv4Addr::from(v4)),
+            IpAddr::V6(v6) => Self::from(std::net::Ipv6Addr::from(v6)),
+        }
+    }
+}
+
 /// `no_std` implementation of `std::net::Ipv4Addr`.
 ///
 /// Note: because we intend to replace this type with `core::net::Ipv4Addr` as soon as it is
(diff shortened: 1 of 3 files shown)


src/server_name.rs
```diff
@@ -49,6 +51,18 @@ impl<'a> ServerName<'a> {
...
+
+    /// Return the string representation of this `ServerName`.
+    #[cfg(all(feature = "alloc", feature = "std"))]
+    pub fn to_str(&self) -> Cow<'_, str> {
+        match self {
+            Self::DnsName(d) => d.as_ref().into(),
+            Self::IpAddress(i) => {
+                let ip: std::net::IpAddr = (*i).into();
+                ip.to_string().into()
+            }
```
cpu, line 63: I think we could make this not require `std` if we wanted to lift the `fmt::Display` impl from core for both the `Ipv4Addr` and `Ipv6Addr` types in this crate. I wasn't sure if it was worth the effort in this case. WDYT?

src/server_name.rs
```diff
@@ -49,6 +51,18 @@ impl<'a> ServerName<'a> {
             Self::IpAddress(i) => ServerName::IpAddress(*i),
         }
     }
+
+    /// Return the string representation of this `ServerName`.
+    #[cfg(all(feature = "alloc", feature = "std"))]
```
djc, line 56: `std` implies `alloc` so we should use `feature = "std"` in favor of `all(feature = "alloc", feature = "std")`.

djc (commented):
It seems to me that `ServerName` is an input-only type from the perspective of webpki/rustls so it's not that obvious why it needs non-trivial implementations to change it back to string form. What's the ffi/platform verifier use case, exactly?

src/server_name.rs
cpu: Fixed throughout.

src/server_name.rs
djc: Probably not worth the effort, the IPv6 one is kinda gnarly.

src/server_name.rs
```diff
@@ -49,6 +51,18 @@ impl<'a> ServerName<'a> {
...
         }
     }
+
+    /// Return the string representation of this `ServerName`.
+    #[cfg(feature = "std")]
+    pub fn to_str(&self) -> Cow<'_, str> {
+        match self {
+            Self::DnsName(d) => d.as_ref().into(),
+            Self::IpAddress(i) => {
+                let ip: std::net::IpAddr = (*i).into();
```
djc, line 61: Can we just make this `std::net::IpAddr::from(*i).to_string().into()`?

src/server_name.rs
```diff
@@ -49,6 +51,18 @@ impl<'a> ServerName<'a> {
             Self::IpAddress(i) => ServerName::IpAddress(*i),
         }
     }
+
+    /// Return the string representation of this `ServerName`.
```
djc, line 55: I think it would be nice to be explicit about the conditions under which this allocates (which are a little unintuitive).

djc (approved):

src/server_name.rs
cpu: Good call, thanks. Fixed up.

src/server_name.rs
cpu: I added some explicit documentation about this. Thanks!

cpu:
Published rustls-pki-types v0.2.3 at registry `crates-io`

cpu:
Argh - I broke semver calling this 0.2.3 :cry: af0db937efa8a3263e3b81d1e3b158d55866f777 was in `main` and added a `Debug` bound on `SignatureVerificationAlgorithm`. 

That makes it incompatible with `rustls-webpki 0.102.0-alpha.7` because `RingAlgorithm` isn't `Debug`.

I could cut a new 0.2.4 that includes this branch's changes but not af0db937efa8a3263e3b81d1e3b158d55866f777, or we could roll forward and fix webpki. I'm not sure what the best course of action is given this crate is pretty internal to the Rustls ecosystem.

cpu:
Here's a webpki fix PR: https://github.com/rustls/webpki/pull/212

cpu:
> I guess we should add a cargo-semver-checks job?

It was there, but CI wasn't running :sob: I should have fixed that before I merged the other work. Silly mistakes.

PR #19, opened 2023-12-05
Allow constructing ServerName from String

Keruspe:
Split the common part out of TryFrom<&str> and allow constructing a ServerName<'static> from a String like we do for DnsName when alloc feature is enabled.

This is useful when the domain name we want to connect to comes from an environmental source


cpu:
>  allow constructing a ServerName<'static> from a String

Isn't this possible with the current API?
```rust
let servername = ServerName::try_from(my_string.as_str()).unwrap().to_owned();
```

Keruspe:
Oh, I totally missed the to_owned, you're right

djc:
That will make an extra allocation though, so it's less efficient?

cpu:
Check out https://github.com/rustls/pki-types/pull/21 and 1.0.1 - it should make this more ergonomic/easier to discover.

Thanks for raising the issue/opening a PR.

PR #20, opened 2023-12-05
server_name: demonstrate using ServerName::to_owned

cpu:
It seems to be a common source of confusion (https://github.com/rustls/pki-types/pull/19, https://github.com/rustls/tokio-rustls/issues/35) that it's possible to construct an owned `ServerName` from an owned `String` using the existing API and `ServerName::to_owned`. This commit adds an explicit example demonstrating this capability to help make it more discoverable.

### DIFF
--- a/src/server_name.rs
+++ b/src/server_name.rs
@@ -29,6 +29,15 @@ use std::error::Error as StdError;
 /// let x = "example.com".try_into().expect("invalid DNS name");
 /// # let _: ServerName = x;
 /// ```
+///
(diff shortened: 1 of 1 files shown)


djc:
I think #21 makes this easier.

cpu:
> I think https://github.com/rustls/pki-types/pull/21 makes this easier.

Agreed.

djc:
(Might still be good to have an example for this? Not sure.)

PR #21, opened 2023-12-05
<ServerName as TryFrom<_>> tweaks

djc:


src/server_name.rs
```diff
@@ -74,6 +74,21 @@ impl<'a> fmt::Debug for ServerName<'a> {
     }
 }
 
+#[cfg(feature = "alloc")]
+impl TryFrom<String> for ServerName<'static> {
```
cpu, line 78: I think we should update the `ServerName` doc block to reference this alongside the existing `ServerName::try_from(&str)`

cpu (approved):

cpu (approved):

src/server_name.rs
djc: Added. Please take a look; feel free to merge and publish if satisfied.

cpu: Thanks, will do

cpu:
* Pushed v/1.0.1 tag
* Published rustls-pki-types v1.0.1 at registry `crates-io`
* Created a [GitHub release](https://github.com/rustls/pki-types/releases/tag/v%2F1.0.1).

PR #22, opened 2023-12-06
Fix server_name docs

rukai:
A missing ``` resulted in incorrectly rendered docs: https://docs.rs/rustls-pki-types/latest/rustls_pki_types/enum.ServerName.html


src/server_name.rs
```diff
@@ -36,8 +36,9 @@ use std::error::Error as StdError;
 ///
 /// which will yield a `ServerName<'static>` if successful.
 ///
-/// // or, alternatively...
 ///
+/// or, alternatively...
```
djc, line 40: Would you mind keeping the empty lines as intended: one before, and one after "or, alternatively..."?

djc (commented):
Thanks for cleaning this up!

cpu (approved):
LGTM. Thanks again for catching/fixing this :+1:

src/server_name.rs
djc: IMO it is more idiomatic to spell this as `let x = ServerName::try_from("example.com").expect();`.

rukai: I mean, thats what the above example is for, right?

djc: That's fair.

djc:
Please squash your commits together so we won't end up with the temporary commits in the commit history.

PR #29, opened 2024-02-07
feat: derive from `Clone` on `PrivateKeyDer` enum

ologbonowiwi:


ctz:
I think we specifically want to avoid making private key material types `Clone`, so it is possible to minimise the number of places they exist in memory. What problem are you trying to solve?

ologbonowiwi:
And also, thanks for the quick response! :smile:

ctz:
These objects are immutable, so having multiple copies is not really useful. Have you considered wrapping them in Rc/Arc? This gives you (~free/cheap) clone without any underlying copies happening.

ologbonowiwi:
It makes sense @ctz! I'm quite new to the Rust ecosystem, so I didn't knew about this "workaround". Thanks for the hand!

ologbonowiwi:
Hey guys, update on my use case: early evaluate the `Item` worked to move it around the application, but it did not work to load the file, as `TlsAcceptor::builder().with_single_cert` needs ownership of the key, which can't be given since it can't be cloned and is inside an `Arc`.

What I'm planning to do to avoid such issue is hold the plain file on the application and late evaluate it to an `Item` right before the creation of `TlsAcceptor`. Anyway, thanks for your kindness and velocity on giving me feedback around this issue.

ologbonowiwi:
It worked @djc, thanks 🙌🏽 was able to achieve a clone-like behavior by using `Arc` + `clone_key`.

I apologize for my ignorance, but what's the difference between deriving explicitly from `Clone` and `clone_key`, and why `clone_key` is safe to use?

djc:
The idea was that `clone_key()` forces an explicit usage, whereas it would be trivial to oversee deriving `Clone` of some type that includes a `PrivateKeyDer`, leading to the key potentially leaking into places it shouldn't go.

It's admittedly a nuanced topic -- this is the line we've ended up at for now.

PR #30, opened 2024-02-07
docs: add mention of no Clone on private keys

cpu:
This comes up with some frequency and since it's a deliberate choice I think it makes sense to call out in the README and in lib.rs's rustdoc header.


ctz (approved):

djc:
No mention of clone_key()?

cpu:
> No mention of clone_key()?

Tried some wording in https://github.com/rustls/pki-types/pull/31 - I'll wait to merge that one until you've had a chance to review.

PR #32, opened 2024-02-10
Supply `web-time::SystemTime` for wasm32

DanGould:
This should be the last piece of logic required for [*rustls* to support WebAssembly](https://github.com/rustls/rustls/issues/808).

`UnixTime::now()` is called for certificate verification in *rustls*. `std::time::SystemTime` is unavailable in wasm32-unknown-unknown, so calling `UnixTime::now()` would panic before this change. [`web-time`](https://github.com/daxpedda/web-time) appears to be a sufficient `std::time` replacement for browser environments. 

Regarding tests, I don't see any calls to `UnixTime::now()` in this crate, so It doesn't seem like testing against a wasm32 would prevent regressions that existing tests would. I'm still happy to set up wasm-bindgen-test if the maintainers would like to see those tests and run them and ci. Perhaps wasm32 tests make sense further up the stack in `rustls` where certificate verification happens and `UnixTime::now()` is called.

I did manually test a rustls connection from wasm32 using this change, however.

I plan to use this to make an https tunnel through a websocket connection in the browser

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,36 @@ jobs:
         env:
           RUST_BACKTRACE: 1
 
+  wasm_build:
+    name: Build wasm32
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Add wasm target
+        run: rustup target add wasm32-unknown-unknown
+
+      - name: wasm32 build (debug; default features)
+        run: cargo build --target wasm32-unknown-unknown --lib
+        env:
+          RUST_BACKTRACE: 1
+
+      - name: wasm32 build (debug; all features)
+        run: cargo build --target wasm32-unknown-unknown --lib --all-features
+        env:
+          RUST_BACKTRACE: 1
+
+      - name: wasm32 build (debug; no default features)
+        run: cargo build --target wasm32-unknown-unknown --lib --no-default-features
+        env:
+          RUST_BACKTRACE: 1
+
   msrv:
     runs-on: ${{ matrix.os }}
     strategy:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,10 @@ categories = ["network-programming", "data-structures", "cryptography"]
 default = ["alloc"]
 alloc = []
 std = ["alloc"]
+web = ["web-time"]
+
+[target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
+web-time = { version = "1", optional = true }
 
 [package.metadata.docs.rs]
 all-features = true
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -38,6 +38,16 @@
 //! [`Rc`]: https://doc.rust-lang.org/std/rc/struct.Rc.html
 //! [`Arc`]: https://doc.rust-lang.org/std/sync/struct.Arc.html
 //! [`PrivateKeyDer::clone_key()`]: https://docs.rs/rustls-pki-types/latest/rustls_pki_types/enum.PrivateKeyDer.html#method.clone_key
+//!
+//! ## Target `wasm32-unknown-unknown` with the `web` feature
+//!
+//! [`std::time::SystemTime`](https://doc.rust-lang.org/std/time/struct.SystemTime.html)
+//! is unavailable in `wasm32-unknown-unknown` targets, so calls to
+//! [`UnixTime::now()`](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.UnixTime.html#method.now),
+//! otherwise enabled by the [`std`](https://docs.rs/crate/rustls-pki-types/latest/features#std) feature,
+//! require building instead with the [`web`](https://docs.rs/crate/rustls-pki-types/latest/features#web)
+//! feature. It gets time by calling [`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now)
+//! in the browser.
 
 #![cfg_attr(not(feature = "std"), no_std)]
 #![warn(unreachable_pub, clippy::use_self)]
@@ -52,8 +62,13 @@ use alloc::vec::Vec;
 use core::fmt;
 use core::ops::Deref;
 use core::time::Duration;
-#[cfg(feature = "std")]
+#[cfg(all(
+    feature = "std",
+    not(all(target_family = "wasm", target_os = "unknown"))
+))]
 use std::time::SystemTime;
+#[cfg(all(target_family = "wasm", target_os = "unknown", feature = "web"))]
+use web_time::SystemTime;
 
 mod server_name;
 pub use server_name::{
@@ -512,7 +527,13 @@ pub struct UnixTime(u64);
 
 impl UnixTime {
     /// The current time, as a `UnixTime`
-    #[cfg(feature = "std")]
+    #[cfg(any(
+        all(
+            feature = "std",
+            not(all(target_family = "wasm", target_os = "unknown"))
+        ),
+        all(target_family = "wasm", target_os = "unknown", feature = "web")
+    ))]
     pub fn now() -> Self {
         Self::since_unix_epoch(
             SystemTime::now()


djc:
Thanks! Can you add a CI job that targets a wasm32 target so that we know that this compiles and doesn't regress?

DanGould:
Added `wasm_build` to build the lib with target wasm32-unknown-unknown in CI

I think this is what you were asking for, but also [made the job run tests on another branch](https://github.com/DanGould/pki-types/tree/wasm-test) in case that was your preference .

ctz:
Not sure about this.

First, it seems quite odd to have to poly-fill bits of std like this. It feels like you actually want to disable the `std` feature in this crate when you use it, and that should drop the availability of the panicking code and require the application to select a replacement. Alternatively, you could stick with `std`, but target a wasm target with a standard syscall interface like wasi (obv. not if you target browsers).

Second, it seems `instant` is not well maintained: https://github.com/sebcrozet/instant/issues/52

djc:
So maybe web-time is a good alternative? @daxpedda do you think it would be a good fit?

Cargo.toml
```diff
@@ -15,6 +15,9 @@ default = ["alloc"]
 alloc = []
 std = ["alloc"]
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
```
daxpedda, line 18:
```suggestion
[target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.dependencies]
```
To make sure this is targeting only `wasm32-unknown-unknown` and not WASI or Emscripten.

daxpedda (changes requested):
> So maybe web-time is a good alternative? @daxpedda do you think it would be a good fit?

AFAICS the goal here is to find a working replacement for `std::time::SystemTime` on the Web platform. In that case `web-time` would be a good fit.

Keep in mind that `wasm32-unknown-unknown` doesn't necessarily mean Web, so if you want to support this target for non-Web platforms, you might want to guard this behind a feature. Obviously this would mean users who target `wasm32-unknown-unknown` without the feature will get a panic when they hit this part of the code.

DanGould:
Thanks so much for the quick response. The most recent force-push should address the concerns raised here. It uses `web-time`, targets only wasm32-unknown-unknown and not WASI or Emscripten, and is guarded behind a new `web` feature.

daxpedda (approved):

djc:
> Obviously this would mean users who target `wasm32-unknown-unknown` without the feature will get a panic when they hit this part of the code.

Could/should we make this a compile time error?

src/lib.rs
```diff
@@ -478,7 +483,10 @@ pub struct UnixTime(u64);
 
 impl UnixTime {
     /// The current time, as a `UnixTime`
-    #[cfg(feature = "std")]
+    #[cfg(any(
+        feature = "std",
+        all(target_family = "wasm", target_os = "unknown", feature = "web")
+    ))]
```
daxpedda, line 489: > > Obviously this would mean users who target `wasm32-unknown-unknown` without the feature will get a panic when they hit this part of the code.
> 
> Could/should we make this a compile time error?

Maybe guarding it would be an option as well:
```suggestion
    #[cfg(any(
        all(feature = "std", not(all(target_family = "wasm", target_os = "unknown"))),
        all(target_family = "wasm", target_os = "unknown", feature = "web")
    ))]
```

DanGould:
Cargo doesn't seem to support `[target.'cfg(all(target_family = "wasm", target_os = "unknown"))'.features]` syntax, so I removed that.

I've also [feature gated `UnixTime`](https://github.com/rustls/pki-types/blob/839d721c202d17f7786c0dceb535a03613e94f9d/src/lib.rs#L484-L499) as suggested by @daxpedda. Those who target wasm32-unknown-unknown without the feature should no longer be able to refer to `UnixTime` and thus should not panic at runtime but get a compile time error instead.

src/lib.rs
DanGould: Are you sure? Isn't it possible to run target wasm32_unknown_unknown in a non-web environment with std feature where std is available?

daxpedda: Calling `SystemTime::now()` will always panic when running it in `wasm32-unknown-unknow`. Unfortunately there is no way around this currently.

See the [Rust Std source code on this](https://github.com/rust-lang/rust/blob/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/sys/wasm/mod.rs#L47-L48).

daxpedda:
> I left the feature gates as-is so that wasm unknown targets running in std-supporting runtimes could use it. Is this something that happens, or do we want this to produce a compile error instead?

For more context: currently `SystemTime::now()` will always panic if you are running under `wasm32-unknown-unknown`. The only way this can change in the future is if a Wasm proposal is introduced that adds instructions to get time. However this is extremely unlikely to happen as this use-case is already covered by WASI, which has its own target.

DanGould:
@daxpedda I took your advice and updated the table above. a wasm32-unknown-unknown target with `std` feature calling `UnixTime::now()` should now produce a compile error from the feature gate.

cpu:
How will this interact with the continuation of the no-std work staged in the Rustls repo. Does the introduction of the `TimeProvider` in https://github.com/rustls/rustls/pull/1502 and https://github.com/rustls/rustls/pull/1688 reduce the need to address this within pki-types, or are the two orthogonal?

daxpedda:
> How will this interact with the continuation of the no-std work staged in the Rustls repo.

I'm not familiar with the `no_std` work in Rustls, but FWIW anything relying on `wasm-bindgen` won't be able to use use `no_std`.

In context for this PR: `wasm32-unknown-unknown` can be `no_std` unless you enable the `web` crate feature.

.github/workflows/ci.yml
```diff
@@ -55,6 +55,38 @@ jobs:
...
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v3
+        with:
+          persist-credentials: false
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: stable
```
cpu, line 70:
```suggestion
        uses: dtolnay/rust-toolchain@stable
```

cpu (commented):
I think the pull request description could use an update (it still refers to using `instant` as opposed to `web-time`).

Given the complexity of the [matrix in this comment](https://github.com/rustls/pki-types/pull/32#issuecomment-1937831405) it also feels like there should be some attempt at documenting the new feature, it scope of support, and its limitations somewhere. Maybe in the top-level `lib.rs` rustdoc?

src/lib.rs
```diff
@@ -38,6 +38,18 @@
...
 //! [`PrivateKeyDer::clone_key()`]: https://docs.rs/rustls-pki-types/latest/rustls_pki_types/enum.PrivateKeyDer.html#method.clone_key
+//!
+//! ## Target `wasm32-unknown-unknown` with the `web` feature
+//!
+//! [`std::time::SystemTime`](https://doc.rust-lang.org/std/time/struct.SystemTime.html)
+//! is unavailable in `wasm32-unknown-unknown` targets, so calls to
+//! [`UnixTime::now()`](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.UnixTime.html#method.now),
+//! otherwise enabled by the [`std`](https://docs.rs/crate/rustls-pki-types/latest/features#std) feature,
+//! require building instead with the [`web`](https://docs.rs/crate/rustls-pki-types/latest/features#web)
+//! feature. It gets time by calling [`Performance.now()`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now).
```
daxpedda, line 49:
```suggestion
//! feature. It gets time by calling [`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now).
```
`Performance.now()` is for `Instant::now()`, `SystemTime::now()` uses `Date.now()`.

src/lib.rs
```diff
@@ -38,6 +38,18 @@
...
+//! in the browser. According to the underlying [`web-time`](https://github.com/daxpedda/web-time)
+//! dependency on which the feature is built, "Emscripten is not supported. WASI doesn't require
+//! support as it has it's own native API to deal with std::time."
```
daxpedda, line 52:
```suggestion
//! dependency on which the feature is built.
```
This is probably wrong to mention here because both Emscripten and WASI are not `wasm32-unknown-unknown`, they have their own separate targets.

FWIW `wasm32-wasi` already supports `std::time`, so nothing needs to be done there. `wasm32-unknown-emscripten`s `std::time` does panic, but AFAIK this can actually be fixed in Std, unlike `wasm32-unknown-unknown`.

daxpedda (changes requested):

djc (approved):

DanGould:
Are any further changes / approval required?

src/lib.rs
```diff
@@ -38,6 +38,16 @@
...
 //! [`PrivateKeyDer::clone_key()`]: https://docs.rs/rustls-pki-types/latest/rustls_pki_types/enum.PrivateKeyDer.html#method.clone_key
+//!
+//! ## Target `wasm32-unknown-unknown` with the `web` feature
+//!
+//! [`std::time::SystemTime`](https://doc.rust-lang.org/std/time/struct.SystemTime.html)
+//! is unavailable in `wasm32-unknown-unknown` targets, so calls to
+//! [`UnixTime::now()`](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.UnixTime.html#method.now),
+//! otherwise enabled by the [`std`](https://docs.rs/crate/rustls-pki-types/latest/features#std) feature,
+//! require building instead with the [`web`](https://docs.rs/crate/rustls-pki-types/latest/features#web)
+//! feature. It gets time by calling [`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now).
```
cpu, line 49: small nit: looks like there's an extra `.` here rendering as "It gets time by calling Date.now()**.** in the browser."


```suggestion
//! feature. It gets time by calling [`Date.now()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now)
```

.github/workflows/ci.yml
```diff
@@ -55,6 +55,36 @@ jobs:
...
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Add wasm target
+        run: rustup target add wasm32-unknown-unknown
+
+       - name: wasm32 build (debug; default features)
+        run: cargo build --target wasm32-unknown-unknown --lib
+        env:
+          RUST_BACKTRACE: 1
```
cpu, line 76: I think the YAML indentation is off here:

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index d87ccfa..1cbaaff 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,10 +70,10 @@ jobs:
       - name: Add wasm target
         run: rustup target add wasm32-unknown-unknown

-       - name: wasm32 build (debug; default features)
+      - name: wasm32 build (debug; default features)
         run: cargo build --target wasm32-unknown-unknown --lib
         env:
-          RUST_BACKTRACE: 1
+           RUST_BACKTRACE: 1

       - name: wasm32 build (debug; all features)
         run: cargo build --target wasm32-unknown-unknown --lib --all-features
```

cpu (approved):
Within my limited knowledge of WASM and the `wasm32-unknown-unknown` this seems reasonable. I flagged one docs nit and a YAML issue that could use quick fixes.

I also I think it'd be nice if the commit adding documentation was folded into the commit adding the feature, but we could probably do that for you before merge if it's a hassle to fix.

DanGould:
adopted @cpu's recommendations and force pushed

.github/workflows/ci.yml
djc: Yeah, I think two-space indentation is expected in YAML?

cpu: You're both right, sorry about that. It was jut the `- name` line that was mis-indented and it looks good now :+1:

cpu (approved):

cpu:
> Not sure about this.

@ctz Does the revised iteration based on `web-time` seem more reasonable to you or is this a non-starter from your perspective no matter the backing dep. filling in for system time on `wasm32-unknown-unknown`?

DanGould:
Regarding @ctz's first concern

> First, it seems quite odd to have to poly-fill bits of std like this. It feels like you actually want to disable the std feature in this crate when you use it, and that should drop the availability of the panicking code and require the application to select a replacement. Alternatively, you could stick with std, but target a wasm target with a standard syscall interface like wasi (obv. not if you target browsers).

This PR was opened so that I can target browsers, which this change alone would allow. It does seem possible to supply web-time by implementing `TimeProvider` further up the stack once https://github.com/rustls/rustls/pull/1502 and https://github.com/rustls/rustls/pull/1688 are merged, but the second is still a draft and I'm not sure on what timeline that would become available.

cpu:
> the second is still a draft and I'm not sure on what timeline that would become available.

I think it will be merged within a week or two (:crossed_fingers:). The main limiting factor is that I'll be out next week.

DanGould:
@cpu https://github.com/rustls/rustls/pull/1688 appears to be paused. Does the presence of that PR make this one obsolete? If that PR is paused and this one allows rustls to be embedded in browser environments, is it worth merging or does it create too much of a maintenance burden?

djc:
If this PR means that web targets can use the default time provider, I think this is probably still a win?

ctz (approved):
I think I'm a bit less concerned now compared to before, given the rather platform-specific nature of this change.

But in the future I think we should resist further polyfills of broken std functionality.

djc (approved):

cpu:
Thanks for sticking with this and iterating based on our feedback @DanGould :+1:

DanGould:
Thanks for the feedback and your patience

guest271314:
Does this provide a means to facilitate the TLS handshake in the browser?

ctz:
I would suggest looking at something like https://github.com/ctz/rustls-embedded-demo -- running rustls in a browser will be a similar job, though obviously the details will be significantly different.

guest271314:
> I would suggest looking at something like https://github.com/ctz/rustls-embedded-demo -- running rustls in a browser will be a similar job, though obviously the details will be significantly different.

I check that out and see what kind of damage I can do with it. Thanks.

PR #35, opened 2024-02-15
fix: fix panic in WASM since std::time::SystemTime::now in used

RRRadicalEdward:
Instead use `time` to get the current time on WASM if new `wasm-bindgen` feature flag is enabled.


CBenoit (approved):

CBenoit:
Sorry about this. The PR was intended to target our own forks. (We are waiting for https://github.com/rustls/pki-types/pull/32 to be merged and published.)

PR #37, opened 2024-02-19
Correct copypasta typos for the pkcs8 and sec1 debug fmt functions.

bmbenson:
Blocks look to have been pasted from PrivatePkcs1KeyDer


djc (approved):
Oops! Thanks for fixing.

PR #39, opened 2024-03-20
Add TryFrom from &[u8] and Vec<u8> to PrivateKeyDer

Alvenix:
I am not very confident about this yet. 

I am mainly using asn1 decoders to learn the differences manually. 

In addition, the approach is heuristic in nature because it would be hard to implement a more complex parser.

I am planning to add more test private key files.


src/lib.rs
```diff
@@ -129,6 +129,60 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
     }
 }
 
+impl<'a> TryFrom<&'a [u8]> for PrivateKeyDer<'a> {
+    type Error = &'static str;
+
+    fn try_from(key: &'a [u8]) -> Result<Self, Self::Error> {
+        if key.len() <= 5 || key[0] != 0x30 {
+            return Err("Invalid key format");
```
djc, line 137: Nit: I think we prefer lowercase error strings.

src/lib.rs
```diff
@@ -129,6 +129,60 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
...
+}
+
+#[cfg(feature = "alloc")]
+impl<'a> TryFrom<Vec<u8>> for PrivateKeyDer<'a> {
+    type Error = &'static str;
+
+    fn try_from(key: Vec<u8>) -> Result<Self, Self::Error> {
+        let private_key_der = PrivateKeyDer::try_from(&key[..])?;
+
+        match private_key_der {
```
djc, line 178: Nit: would be nice to hoist the `Ok()` out of the `match`.

src/lib.rs
```diff
@@ -129,6 +129,60 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
...
+        }
+    }
+}
+
+#[cfg(feature = "alloc")]
+impl<'a> TryFrom<Vec<u8>> for PrivateKeyDer<'a> {
+    type Error = &'static str;
+
+    fn try_from(key: Vec<u8>) -> Result<Self, Self::Error> {
+        let private_key_der = PrivateKeyDer::try_from(&key[..])?;
```
djc, line 176: Nit: I think it would be okay to just inline the binding here? (As in, `match PrivateKeyDer::try_from(&key[..])? { .. }`.)

djc (commented):
Looks pretty good, thanks for picking this up!

I'll let @cpu and @ctz judge the specifics of the DER parsing.

src/lib.rs
```diff
@@ -129,6 +129,60 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
...
+
+        // We skip the tag and the length of the sequence
+        let mut skip_len = 2;
+
+        // We skip additonal length bytes
+        if key[1] >= 0x81 {
+            skip_len += key[1] - 0x80;
+        }
+
+        match key[skip_len as usize..] {
```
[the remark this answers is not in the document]
djc: > If it is okay to add a more complex test, I would like to use openssl to generate large number of private keys programmatically and ensure the function chooses a correct key type.

Depends on the complexity? I think several hundred lines of code would be worth it if it doesn't blow up test cycle times beyond a few minutes.

src/lib.rs
```diff
@@ -129,6 +129,58 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
...
+
+            Some([0x02, 0x01, 0x01, // This is decoded as PKI header with value 1
+            ..]) => {
+                Ok(Self::Sec1(key.into()))
+            }
+
+            Some([0x02, 0x01, 0x00, // Version = 0
+            ..]) => {
+                Ok(Self::Pkcs8(key.into()))
+            }
```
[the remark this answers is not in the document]
ctz: There are some suitable files in https://github.com/rustls/rustls/tree/main/rustls/src/testdata (rsa*.der, eddsakey.der, nist*.der)

src/lib.rs
```diff
@@ -129,6 +129,57 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
...
+    fn try_from(key: &'a [u8]) -> Result<Self, Self::Error> {
+        if key.len() <= 5 || key[0] != 0x30 {
+            return Err("invalid key format");
+        }
+
+        // We skip the tag and the length of the sequence
+        let mut skip_len = 2;
+
+        // We skip additonal length bytes
+        if key[1] >= 0x81 {
```
cpu, line 144: I think this is off-by-one no? The maximum value for a short-form DER encoded length is 128 (0x80).

cpu (commented):
I think the general approach here is workable, but I'd like to see it written with less magic values and more information about the heuristics in use and why we think they're stable signals.

In reviewing the diff I adapted your code to be closer to what I'm thinking of. Rather than deliver a bunch of review feedback I opened a second PR based on your own with what I'm thinking: https://github.com/rustls/pki-types/pull/40

WDYT?

src/lib.rs
cpu: Possible I've confused myself here :-)

>  The maximum value for a short-form DER encoded length is 128 (0x80).

I think I should have said 127 here. I believe the short form encoded length values that are possible are `[0 ... 128)` in range notation.

> I don't think it is off-by-one. If the value is 128 or less we would only need to skip two bytes as assigned in line 141.

I think a value of 128 exactly would be `b1000_0000`, and the leading 1 bit indicates it's not a short form length.

cpu: > I think a value of 128 exactly would be b1000_0000, and the leading 1 bit indicates it's not a short form length.

but... in practice your code would treat it as a long form length with 0 additional bytes, skipping 2 total. So I think it's a moot point in this context anyway.

It doesn't matter how many times I look at this stuff I always find it tricky...

cpu: I updated the replacement PR's desc to point to this thread and clarify it wasn't really an off-by-one. :+1:

PR #40, opened 2024-03-21
Add TryFrom from &[u8] and Vec<u8> to PrivateKeyDer

cpu:
This branch is an alternative to https://github.com/rustls/pki-types/pull/39 that takes the base commit and reworks it slightly to:

* Introduce constants for the various bits of ASN.1 DER we're using, closely matching defines found in the webpki crate's DER module.
* Simplify some of the bounds checking and indexing.
* ~Correct the boundary for the short form length check.~ (I think we've reasoned out [this was a no-op](https://github.com/rustls/pki-types/pull/39#discussion_r1534262298) change)
* Document the heuristics in use, and why they work, citing RFCs for the ASN.1 structures we're matching.

Along the way I also:
* Used `cargo fuzz` to initialize a fuzzing crate, and a target for fuzzing this new API surface.
* Added a fuzzing smoke-test target to CI for the above
* Simplified the included unit test to use a table-driven design

I thought that it would be easier to open this PR rather than deliver a large volume of review feedback.  Many thanks to @Alvenix for the original implementation.


cpu:
> cpu marked this pull request as draft now

I'm using something that requires std. Need to adjust :face_exhaling:

cpu:
>  Why couldn't the tests be moved with the other tests if they need std?

I think the issue isn't in the tests, it's the way I'm trying to build a pattern to match for the `pkcs8_hint` by combining two `&[u8]` at runtime with `concat`, creating a `Vec`. 

There isn't a nice way to make a compile-time const that combines two const slice fragments without defining a macro (as far as I can tell). I'll have to rework the approach slightly.

src/lib.rs
```diff
@@ -129,6 +129,107 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
     }
 }
 
+impl<'a> TryFrom<&'a [u8]> for PrivateKeyDer<'a> {
+    type Error = &'static str;
+
+    fn try_from(key: &'a [u8]) -> Result<Self, Self::Error> {
+        const INVALID_ERR: &str = "unknown or invalid key format";
```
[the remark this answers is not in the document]
cpu: I moved this to a `static` :+1:

cpu:
> There isn't a nice way to make a compile-time const that combines two const slice fragments without defining a macro (as far as I can tell). I'll have to rework the approach slightly.

I gave up on pulling out the VERSION_ONE and VERSION_TWO slices and inlined their values into the three hints. It's a little less nice but I think making a macro would be overkill.

cpu:
> Used cargo fuzz to initialize a fuzzing crate, and a target for fuzzing this new API surface.

FWIW I ran the fuzzer "a while" (~30m) on this branch, and the original commit from #39 without any findings.

src/lib.rs
```diff
@@ -129,6 +129,106 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
...
+        //      attributes [0] Attributes OPTIONAL
+        //   }
+        // PKCS#5 (https://www.rfc-editor.org/rfc/rfc8018) describes the AlgorithmIdentifier
+        // as a SEQUENCE.
+        //
+        // Therefore, we consider the outer SEQUENCE, a Version of 0, and the start of
+        // an AlgorithmIdentifier to be enough to identify a PKCS#8 key. If it were PKCS#1,
+        // the version would not be followed by a SEQUENCE. If it were SEC1, the version would
+        // not have been 0.
+        let pkcs8_hint = [TAG_INTEGER, 0x01, 0x00, TAG_SEQUENCE];
```
djc, line 171: All things considered, I would probably structure this code to keep the hint definitions and their usage closer together. So maybe just:

```rust
if key_bytes.starts_with([TAG_INTEGER, 0x01, 0x00, TAG_SEQUENCE]) {
     return Ok(Self::Pkcs8(key.into());
}
```

And similar for the other variants. This keeps the documentation and all of the related code closer together, which IMO will make this easier to follow.

src/lib.rs
```diff
@@ -129,6 +129,106 @@ impl<'a> From<PrivatePkcs8KeyDer<'a>> for PrivateKeyDer<'a> {
...
+}
+
+static INVALID_KEY_DER_ERR: &str = "unknown or invalid key format";
+
+#[cfg(feature = "alloc")]
+impl<'a> TryFrom<Vec<u8>> for PrivateKeyDer<'a> {
+    type Error = &'static str;
+
+    fn try_from(key: Vec<u8>) -> Result<Self, Self::Error> {
+        Ok(match PrivateKeyDer::try_from(&key[..])? {
```
djc, line 224: Is it on purpose that you're using `PrivateKeyDer` here instead of `Self`? It seems a little weird when `Self` is `PrivateKeyDer` to use both (but maybe this is one of those cases where the lifetime thing interferes with using `Self`).

djc (approved):
This looks great!

In terms of commit history, I would prefer not to keep the original @Alvenix commit separate from the other commits that change the code/tests, and instead squash them together (and use Co-authored-by or whatever the tag is to clarify origination) -- keeping the fuzzing commit separate is fine, of course.

src/lib.rs
cpu: full disclosure: I took this part from the original PR without much thought :-)

>  maybe this is one of those cases where the lifetime thing interferes with using Self

I believe that's the case. Making a quick swap I get a `key` does not live long enough error.

src/lib.rs
cpu: I implemented this feedback and agree it's much easier to follow :+1:

cpu:
> In terms of commit history, I would prefer not to keep @Alvenix' original commit separate from the other commits that change the code/tests, and instead squash them together (and use Co-authored-by or whatever the tag is to clarify origination)

That sounds good to me. I've pushed an update that folds the revisions and the test updates into the original commit, maintaining the authorship from #39. I left the fuzzing commits separate.

djc (commented):
💯

ctz (approved):

PR #44, opened 2024-04-07
Move test keys into tests

djc:
IMO this is a more appropriate/conventional home for test data (compared to inside `src`).

(Context: looking at https://github.com/rustls/rustls/pull/1462 again.)


ctz (approved):
I feel like behaviour tests that strictly use the public API should also not be in `src/` (rationale: it's easy to accidentally neglect to export a public API)

cpu (approved):
> I feel like behaviour tests that strictly use the public API should also not be in src/

Agreed. Moving some of the existing tests seems like a good follow-up.

djc:
I'll follow up with those changes.

PR #45, opened 2024-04-08
Move key type determination tests into tests/

djc:
As discussed in #44.


ctz (approved):
👍🏽

cpu (approved):
Thanks. I like the extra tidying :+1:

PR #46, opened 2024-04-10
add EchConfigListBytes for encrypted client hello configs

cpu:
### lib: DerInner -> BytesInner
The crate-internal `DerInner` type is really a `Cow`-like type for generic bytes. The nature of those bytes is indicated by the pub wrapper type holding the `DerInner`.

With that in mind we can re-use it for other non-DER types (e.g. ECH configs) but first need to rename the type to `BytesInner` since it is no longer DER in all cases.

### EchConfigListBytes for encrypted client hello configs

Raw ECH config lists are likely to be shared between Rustls and other crates (e.g. they may be read from `rustls-pemfile` and `.pem` inputs for server configuration, or they may be fetched from DNS using DNS-over-HTTPS using `hickory-dns` for client config).

This commit introduces a `EchConfigListBytes` type, wrapping `BytesInner`, that allows representing a borrowed or owned raw TLS encoded ECH config list, corresponding to the `ECHConfigList` type specified in TLS presentation language in [draft-ietf-tls-esni-18 §4](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-18#section-4). This type can be used between crates to maintain typed context for the bytes throughout.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls-pki-types"
-version = "1.4.1"
+version = "1.5.0"
 edition = "2021"
 rust-version = "1.60"
 license = "MIT OR Apache-2.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -249,14 +249,14 @@ impl PrivatePkcs1KeyDer<'_> {
 
 impl<'a> From<&'a [u8]> for PrivatePkcs1KeyDer<'a> {
     fn from(slice: &'a [u8]) -> Self {
-        Self(Der(DerInner::Borrowed(slice)))
+        Self(Der(BytesInner::Borrowed(slice)))
     }
 }
 
 #[cfg(feature = "alloc")]
 impl<'a> From<Vec<u8>> for PrivatePkcs1KeyDer<'a> {
     fn from(vec: Vec<u8>) -> Self {
-        Self(Der(DerInner::Owned(vec)))
+        Self(Der(BytesInner::Owned(vec)))
     }
 }
 
@@ -291,14 +291,14 @@ impl PrivateSec1KeyDer<'_> {
 
 impl<'a> From<&'a [u8]> for PrivateSec1KeyDer<'a> {
     fn from(slice: &'a [u8]) -> Self {
-        Self(Der(DerInner::Borrowed(slice)))
+        Self(Der(BytesInner::Borrowed(slice)))
     }
 }
 
 #[cfg(feature = "alloc")]
 impl<'a> From<Vec<u8>> for PrivateSec1KeyDer<'a> {
     fn from(vec: Vec<u8>) -> Self {
-        Self(Der(DerInner::Owned(vec)))
+        Self(Der(BytesInner::Owned(vec)))
     }
 }
 
@@ -333,14 +333,14 @@ impl PrivatePkcs8KeyDer<'_> {
 
 impl<'a> From<&'a [u8]> for PrivatePkcs8KeyDer<'a> {
     fn from(slice: &'a [u8]) -> Self {
-        Self(Der(DerInner::Borrowed(slice)))
+        Self(Der(BytesInner::Borrowed(slice)))
     }
 }
 
 #[cfg(feature = "alloc")]
 impl<'a> From<Vec<u8>> for PrivatePkcs8KeyDer<'a> {
     fn from(vec: Vec<u8>) -> Self {
-        Self(Der(DerInner::Owned(vec)))
+        Self(Der(BytesInner::Owned(vec)))
     }
 }
 
@@ -658,12 +658,12 @@ impl UnixTime {
 /// the data is owned (by a `Vec<u8>`) or borrowed (by a `&[u8]`). Support for the owned
 /// variant is only available when the `alloc` feature is enabled.
 #[derive(Clone)]
-pub struct Der<'a>(DerInner<'a>);
+pub struct Der<'a>(BytesInner<'a>);
 
 impl<'a> Der<'a> {
     /// A const constructor to create a `Der` from a borrowed slice
     pub const fn from_slice(der: &'a [u8]) -> Self {
-        Self(DerInner::Borrowed(der))
+        Self(BytesInner::Borrowed(der))
     }
 }
 
@@ -671,8 +671,8 @@ impl AsRef<[u8]> for Der<'_> {
     fn as_ref(&self) -> &[u8] {
         match &self.0 {
             #[cfg(feature = "alloc")]
-            DerInner::Owned(vec) => vec.as_ref(),
-            DerInner::Borrowed(slice) => slice,
+            BytesInner::Owned(vec) => vec.as_ref(),
+            BytesInner::Borrowed(slice) => slice,
         }
     }
 }
@@ -687,14 +687,14 @@ impl Deref for Der<'_> {
 
 impl<'a> From<&'a [u8]> for Der<'a> {
     fn from(slice: &'a [u8]) -> Self {
-        Self(DerInner::Borrowed(slice))
+        Self(BytesInner::Borrowed(slice))
     }
 }
 
 #[cfg(feature = "alloc")]
 impl From<Vec<u8>> for Der<'static> {
     fn from(vec: Vec<u8>) -> Self {
-        Self(DerInner::Owned(vec))
+        Self(BytesInner::Owned(vec))
     }
 }
 
@@ -713,16 +713,16 @@ impl PartialEq for Der<'_> {
 impl Eq for Der<'_> {}
 
 #[derive(Clone)]
-enum DerInner<'a> {
+enum BytesInner<'a> {
     #[cfg(feature = "alloc")]
     Owned(Vec<u8>),
     Borrowed(&'a [u8]),
 }
 
 #[cfg(feature = "alloc")]
-impl DerInner<'_> {
-    fn into_owned(self) -> DerInner<'static> {
-        DerInner::Owned(match self {
+impl BytesInner<'_> {
+    fn into_owned(self) -> BytesInner<'static> {
+        BytesInner::Owned(match self {
             Self::Owned(vec) => vec,
             Self::Borrowed(slice) => slice.to_vec(),
         })
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -502,7 +502,7 @@ impl CertificateDer<'_> {
 
 /// A TLS-encoded Encrypted Client Hello (ECH) configuration list (`ECHConfigList`); as specified in
 /// [draft-ietf-tls-esni-18 §4](https://datatracker.ietf.org/doc/html/draft-ietf-tls-esni-18#section-4)
-#[derive(Clone)]
+#[derive(Clone, Eq, PartialEq)]
 pub struct EchConfigListBytes<'a>(BytesInner<'a>);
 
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -729,6 +771,86 @@ impl DerInner<'_> {
...
+/// Data encoded according to TLS presentation language, either owned or borrowed.
+///
+/// This wrapper type is used to represent TLS-encoded data in a way that is agnostic to whether
+/// the data is owned (by a `Vec<u8>`) or borrowed (by a `&[u8]`). Support for the owned
+/// variant is only available when the `alloc` feature is enabled.
+///
+/// For more information on "TLS encoding" or the TLS presentation language, see
+/// [RFC 8446 §3](https://www.rfc-editor.org/rfc/rfc8446#section-3).
+#[derive(Clone)]
+pub struct TlsBytes<'a>(TlsBytesInner<'a>);
```
djc, line 783: How do we feel about renaming `Der` instead to something more generic, so that we don't have to duplicate all this stuff? We can retain `Der` as a type alias for compatibility.

(The next question is bound to be, what do we call it -- `ByteSlice`, `Cow`, `Bytes`? I don't like `Bytes` because it is already a widely used type; `Cow` is too, and also a bit of a misnomer because we're largely treating it as immutable so the copy-on-write behavior isn't so relevant.)

ctz: Unless I've overlooked something, we could:

- rename the existing `DerInner` type to make it more generic (to some extent the name doesn't matter, because it doesn't leak out the crate -- `BytesInner`? just `Inner`?)
- directly use that type inside `EchConfigBytes`, therefore eliminating the whole `TlsBytes` type? ( A later `Sct` type could take the same approach, and wouldn't require a new intermediate type like `TlsBytes`.)

It seems like `EchConfigBytes` already must have the totality of its interface, because its interior `TlsBytes` is non-pub. I suppose the question is whether the `TlsBytes` type is expected to be useful separately?

(The same argument could be made for the `Der` type, but that's more academic since it's in released versions.)

cpu: These ideas sound reasonable. My main concern was avoiding a semver break, but I think the proposals above would avoid that.

I don't think the `TlsBytes` type is likely to be super useful on its own. I was mainly pattern matching the existing approach, which as you mention, is perhaps of questionable utility w.r.t `Der` in the public interface.

I'll adjust w/ the proposal Ctz put forward,  `DerInner` -> `BytesInner` and `EchConfigBytes(TlsBytes)` -> `EchConfigBytes(BytesInner)`.

djc: Yup, sounds good to me.

cpu: Done. :ballot_box_with_check:

cpu:
> Cargo: v1.4.1 -> v1.5.1

I tacked on a version bump.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rustls-pki-types"
-version = "1.4.1"
+version = "1.5.1"
```
djc, line 3: Should become 1.5.0 I think?

djc (commented):
Is PEM encoding of EchConfig a common/described use case somewhere, or something you're hypothesizing might be useful?

djc (approved):

Cargo.toml
cpu: Oops, yes. Thank you.

cpu:
> Is PEM encoding of EchConfig a common/described use case somewhere, or something you're hypothesizing might be useful?

It's a real use case but there's not a consistent implementation across the ecosystem yet.

* In BoringSSL land their helper tools (e.g. `bssl generate-ech`) spit out and consume binary files. I don't think they've implemented anything for PEM.
* For OpenSSL, and in the standards space, there's a [draft](https://www.ietf.org/archive/id/draft-farrell-tls-pemesni-06.html) from the author of the WIP OpenSSL support for ECH (https://github.com/openssl/openssl/pull/22938). The described `.pem` format is implemented and used there. It's specific to the server usecase and prescribes both a PKCS#8 private key and a single ECHConfig per .pem
* For Cloudflare's Go fork they're using [their own thing](https://github.com/cloudflare/go/blob/ec0a014545f180b0c74dfd687698657a9e86e310/src/crypto/tls/ech_test.go#L103-L117) that isn't specified anywhere that I know of, `-----BEGIN ECH KEYS-----` for keys (plural) and `-----BEGIN ECH CONFIGS-----` for ECH configs (plural). 

Of the set I think `draft-farrell-tls-pemesni-06` looks useful but is a bit rough and early days. I'd rather not implement something with zero spec ala CF's impl. Not implementing anything and using binary formats is OK short term, but I think the proliferation of `.pem` for other things shows there's demand for a non-binary format. I have a branch of pemfile implementing it but I'm not sure its worth landing yet, except perhaps to try and move `draft-farrell-tls-pemesni-06` forward and get more people to converge on that format.

cpu:
> I have a branch of pemfile implementing it but I'm not sure its worth landing yet,

I was thinking something like this: https://github.com/rustls/pemfile/commit/3d3a9ef6d30fddfc6bf8ab4f42ffd70e3964a492

cpu:
> cpu marked this pull request as draft now

I want to hold on to this a bit longer. I might want to add another type for a TLS encoded list of configs.

cpu:
> I want to hold on to this a bit longer. I might want to add another type for a TLS encoded list of configs.

yeah I think this needs some rework, the new type should be for representing the encoding of a **list**, specifically `ECHConfig ECHConfigList<1..2^16-1>` from the draft, not a `ECHConfig` as implied by this branch. 

The encoding of a single item is equivalent to a list of size 1. The raw value read from DNS that we want to represent here is in fact a list, I've just been confused by the common case being a single item that doesn't appear to be a list due to the rules of TLS presentation syntax.

cpu:
> I've just been confused by the common case being a single item that doesn't appear to be a list due to the rules of TLS presentation syntax.

There was another aspect to my confusion I figured out over the weekend. The Hickory DNS representation of `EchConfig` Svc param values [strips the list length prefix](https://github.com/hickory-dns/hickory-dns/blob/94ac564c3f677e038f7255ddb762e9301d0f2c5d/crates/proto/src/rr/rdata/svcb.rs#L830-L839), storing only the raw bytes after that in its wrapped `Vec`. The docs don't describe what format this `Vec` should be considered as holding, but it isn't the wire format (it's missing the length prefix required) and it isn't the presentation format (it's not the base64 of the wire format). You can get the value with the list prefix by calling [`EchConfig.emit()`](https://docs.rs/hickory-proto/latest/hickory_proto/serialize/binary/trait.BinEncodable.html#tymethod.emit) with a `BinEncoder`. I will do that in my client code as a temporary workaround to be able to treat everything consistently.

I think this is the wrong way for Hickory to present this data to a consumer.  I believe most users of the data are going to be expecting the wire format described in the `draft-ietf-tls-svcb-ech-01` and `draft-ietf-tls-esni-18`. It's easier to deserialize the list when you know can treat it the same as any other TLS encoded list, but this requires the list length prefix be in place. I'll propose an upstream change shortly to rework this interface.

In the meantime I think my conclusion is that we should **not** have a pki-types representation for a single `ECHConfig`, but only for `ECHConfigList`. That's what [draft-ietf-tls-svcb-ech-01](https://datatracker.ietf.org/doc/html/draft-ietf-tls-svcb-ech-01) describes as going in the DNS. That's what our client facing code wants to load to pick a compatible ECH config. That's what servers will want to load to specify the configs it understands/offers for retries.

cpu:
> I'll propose an upstream change shortly to rework this interface.

https://github.com/hickory-dns/hickory-dns/pull/2183

cpu:
> In the meantime I think my conclusion is that we should not have a pki-types representation for a single ECHConfig, but only for ECHConfigList.

I've reworked the branch around this conclusion and updated the commit/PR description accordingly. I think this is ready for re-review & I'll have an update to the WIP Rustls ECH PR to take a git patch on this branch shortly.

While I was here I also tacked on the ubuntu-20.04 -> ubuntu-latest change I'm proposing in https://github.com/rustls/webpki/pull/248.

src/lib.rs
```diff
@@ -500,6 +500,64 @@ impl CertificateDer<'_> {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        hex(f, self.as_ref())
+    }
+}
+
+impl PartialEq for EchConfigListBytes<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        self.as_ref().eq(other.as_ref())
+    }
+}
```
djc, line 526: Why can't we derive this? (Same for `Eq`.)

djc (approved):

src/lib.rs
cpu: There's no `Eq`/`PartialEq` implemented for `BytesInner` (Prev `DerInner`). The pre-existing `Der` type that also wraps `BytesInner` is implementing `Eq`/`PartialEq` by hand as well as a result. 

I think it would be reasonable to implement `PartialEq` for `BytesInner` (unless there's a reason you avoided it when adding `DerInner` I'm missing?) - it's a crate-internal type so I don't think we need to worry about folks comparing a `BytesInner` from a `Der` with a `BytesInner` from a `EchConfigListBytes`. 

I'll tack on a commit changing this up for both `Der` and `EchConfigListBytes`.

cpu: > I'll tack on a commit changing this up for both Der and EchConfigListBytes.

Done.

djc:
Can we derive them for BytesInner then, instead of manually implementing?

djc:
Ah, yes!

cpu:
> Wouldn't a derived impl consider the enum variants first, and then their contents?

Yeah, I think so:
```rust
    #[test]
    fn bytes_inner_equality() {
        let owned = BytesInner::Owned(vec![1, 2, 3]);
        let borrowed = BytesInner::Borrowed(&[1, 2, 3]);
        assert_eq!(owned, borrowed);
    }
```

With the derived impl:
```
thread 'tests::bytes_inner_equality' panicked at src/lib.rs:825:9:
assertion `left == right` failed
  left: Owned([1, 2, 3])
 right: Borrowed([1, 2, 3])
```

With my manual impl:
```
test tests::bytes_inner_equality ... ok
```

I'll add some test coverage for this while I'm here. The derived version passed without error before I added a test.

djc:
Can we do `self.as_ref() == other.as_ref()` in the `PartialEq` impl?

cpu:
> Can we do self.as_ref() == other.as_ref() in the PartialEq impl?

Ah! good idea. It's was a similar situation here: `AsRef` was being implemented for the wrapping types manually, not for `BytesInner`. I dragged that trait impl down, used it in the partial eq, and updated the manual impls on the wrappers to defer to the wrapped type impl.

djc (approved):

cpu:
@ctz Do you have interest in reviewing this branch? I think its in a stable state now and I don't expect to need any other ECH related changes in this repo.

cpu:
* admin merged
* fixed the branch protection rules for `ubuntu-latest`.
* pushed [v/1.5.0](https://github.com/rustls/pki-types/releases/tag/v%2F1.5.0) tag
* Published rustls-pki-types v1.5.0 at registry `crates-io`
* Created [v1.5.0 GitHub release](https://github.com/rustls/pki-types/releases/tag/v%2F1.5.0)

PR #50, opened 2024-06-14
Fix some new 1.79 clippy findings

cpu:
This branch fixes two new classes of clippy findings from the [just-released 1.79 toolchain](https://blog.rust-lang.org/2024/06/13/Rust-1.79.0.html).


djc (approved):

cpu:
I did a quick pass through the other repo scheduled CI runs from yesterday and it looks like this was the only one that had 1.79 lint findings :+1:

PR #53, opened 2024-09-03
Direct support for PEM-decoding of this crate's types

ctz:
This incorporates the contents of the pemfile crate, including https://github.com/rustls/pemfile/pull/53

https://github.com/rustls/pemfile/pull/55 is a draft over there, to rebase that crate on top of this.


src/base64.rs
```diff
@@ -0,0 +1,730 @@
+/// Provide an upper limit on how much space could be required
```
djc, line 1: I see you just added these files here without integrating them in the crate (via `mod` declarations), which you deferred to the next commit (at least for one of the modules?). This all seems a little iffy in terms of our usual goal of having each commit pass CI... Since IIRC `base64` hasn't made it into the committed `main` history of the pemfile repo, I wouldn't worry too much about this separate step.

So I think I would suggest squashing the first three commits together -- I don't think there are substantial changes here that are worth keeping separate.

src/pem.rs
```diff
@@ -57,6 +57,7 @@ pub enum Item {
 
 /// Errors that may arise when parsing the contents of a PEM file
 #[derive(Debug, PartialEq)]
+#[non_exhaustive]
```
djc, line 60: Tiny nit: I think we usually put `#[non_exhaustive]` before the derives?

src/lib.rs
```diff
@@ -757,6 +807,53 @@ impl UnixTime {
...
 }
 
+/// An extension trait for types we can decode from PEM.
+#[cfg(feature = "alloc")]
+pub trait DecodePem {
+    /// Underlying function to be implemented by target types.
+    ///
+    /// This is not intended for direct use, instead use `decode_from_pem`
+    /// and other provided functions in this trait.
+    fn from_pem_items(
```
djc, line 817: Suggest moving this below the other methods.

src/lib.rs
```diff
@@ -757,6 +807,53 @@ impl UnixTime {
...
+    fn decode_from_pem_str(pem: &str) -> Result<Self, pem::Error>
+    where
+        Self: Sized,
+    {
+        Self::decode_from_pem(pem.as_bytes())
+    }
+
+    /// Decode this type from PEM present in a buffered reader.
+    #[cfg(feature = "std")]
+    fn decode_from_pem_reader(rd: &mut impl std::io::BufRead) -> Result<Self, pem::IoError>
```
djc, line 841: Nit: suggest skipping the `decode_` prefix on these.

src/pem.rs
```diff
@@ -269,3 +272,29 @@ fn read_until_newline<R: io::BufRead + ?Sized>(r: &mut R, buf: &mut Vec<u8>) ->
 pub fn read_all(rd: &mut dyn io::BufRead) -> impl Iterator<Item = Result<Item, IoError>> + '_ {
     iter::from_fn(move || read_one(rd).transpose())
 }
+
+/// Iterate over all PEM sections by reading `pem_slice`
+pub(crate) fn read_all_from_slice(
```
djc, line 277: Ordering nit: before `read_one_from_slice()`?

src/lib.rs
```diff
@@ -120,6 +127,25 @@ impl<'a> PrivateKeyDer<'a> {
     }
 }
 
+#[cfg(feature = "alloc")]
+impl DecodePem for PrivateKeyDer<'static> {
+    fn from_pem_items(
+        iter: &mut impl Iterator<Item = Result<pem::Item, pem::Error>>,
```
djc, line 133: Nit: maybe import `Item` at the module level?

src/lib.rs
```diff
@@ -757,6 +807,53 @@ impl UnixTime {
...
+    ///
+    /// This is not intended for direct use, instead use `decode_from_pem`
+    /// and other provided functions in this trait.
+    fn from_pem_items(
+        iter: &mut impl Iterator<Item = Result<pem::Item, pem::Error>>,
+    ) -> Result<Self, pem::Error>
+    where
+        Self: Sized;
+
+    /// Decode this type from PEM contained a byte slice.
```
djc, line 823: "contained [in] a"?

src/lib.rs
```diff
@@ -757,6 +807,53 @@ impl UnixTime {
     }
 }
 
+/// An extension trait for types we can decode from PEM.
+#[cfg(feature = "alloc")]
+pub trait DecodePem {
```
djc, line 812: Use `: Sized` as a supertrait so we don't need to repeat it on every method?

src/lib.rs
```diff
@@ -19,7 +19,14 @@
...
 //!
-//! The [rustls-pemfile](https://docs.rs/rustls-pemfile) crate can be used to parse PEM files.
+//! Types here can be created from:
+//!
+//! - DER using (for example) [`PrivatePkcs8KeyDer::from`].
+//! - PEM using (for example) [`PrivatePkcs8KeyDer::decode_from_pem`] via the [`DecodePem`] extension trait.
+//!
+//! `decode_from_pem` returns the first matching item from the given input.
+//! It is usual for given PEM file to contain multiple items: if you wish
+//! to examine all of these you can use the iterator-based API in the [`pem`] module.
```
djc, line 29: I think we should have a convenience API that fits the use case of easily getting a `Vec<CertificateDer<'static>>` from a file, and this shouldn't require going via the `Item` API.

ctz: Does `Vec<CertificateDer<'static>>::decode_from_pem_reader(BufReader::new(File::open("certs.pem")?))` (modulo name of that function) meet that?

Maybe we could add a decode-from-io::Read, reducing that to `Vec<CertificateDer<'static>>::decode_from_pem_reader(&mut File::open("certs.pem")?)` and hiding the BufReader requirement ?

djc: > Does `Vec<CertificateDer<'static>>::decode_from_pem_reader(BufReader::new(File::open("certs.pem")?))` (modulo name of that function) meet that?

I feel like writing anything like `Vec<FooBar>::method()` is pretty unusual to write/read, though I guess `Vec::from_pem_reader()` would work as long as there is type inference? Might be fine.

> Maybe we could add a decode-from-io::Read, reducing that to `Vec<CertificateDer<'static>>::decode_from_pem_reader(&mut File::open("certs.pem")?)` and hiding the BufReader requirement ?

Paving the cow path for `File` would be nice, I think.

src/pem.rs
```diff
@@ -0,0 +1,280 @@
...
+    Base64Decode(String),
+}
+
+/// Extract and decode the next PEM section from `input`
+///
+/// - `Ok(None)` is returned if there is no PEM section to read from `input`
+/// - Syntax errors and decoding errors produce a `Err(...)`
+/// - Otherwise each decoded section is returned with a `Ok(Some((Item::..., remainder)))` where
+///   `remainder` is the part of the `input` that follows the returned section
+pub fn read_one_from_slice(mut input: &[u8]) -> Result<Option<(Item, &[u8])>, Error> {
```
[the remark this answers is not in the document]
ctz: I think I was trying to leave this module somewhat unchanged API-wise, so the final rustls-pemfile 2.x release can just reexport the same items as before from rustls-pki-types?

But, that does mean we're committing to the same API in rustls-pki-types 1.x, in exchange for avoiding a rustls-pemfile 3.x, or a more complex translation layer in rustls-pemfile 2.x?

src/base64.rs
```diff
@@ -0,0 +1,730 @@
...
+/// to decode a base64 encoding of len `base64_len`
+pub(crate) fn decoded_length(base64_len: usize) -> usize {
```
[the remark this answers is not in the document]
ctz: moved below other pub(crate) members

src/pem.rs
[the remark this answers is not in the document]
djc: I was thinking a slightly more complex translation layer in rustls-pemfile 2.x might be the better option, especially if we can avoid duplicating function bodies and (non-trivial) data types.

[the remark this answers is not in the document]
ctz: Sounds good to me! I will maintain the matching pemfile changes on https://github.com/rustls/pemfile/tree/jbp-use-pki-types-pem

src/pem.rs
```diff
@@ -74,6 +74,17 @@ pub enum Error {
     Base64Decode(String),
 }
 
+/// Errors that may arise from reading from a file-like stream
+#[cfg(feature = "std")]
+#[derive(Debug)]
+pub enum IoError {
```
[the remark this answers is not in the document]
ctz: Now we've decided to make some breaking changes here, I will add this as a new `Error` variant.

src/pem.rs
```diff
@@ -0,0 +1,293 @@
...
+    if let Some((section_type, end_marker)) = section.as_ref() {
+        if line.starts_with(end_marker) {
+            let mut der = vec![0u8; base64::decoded_length(b64buf.len())];
+
+            let der_len = match &section_type[..] {
+                b"RSA PRIVATE KEY" | b"PRIVATE KEY" | b"EC PRIVATE KEY" => {
+                    base64::decode_secret(b64buf, &mut der)
+                }
+                _ => base64::decode_public(b64buf, &mut der),
+            }
```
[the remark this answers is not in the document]
cpu: > WDYT about a direction like this? https://github.com/rustls/pki-types/commit/32be21d18276b5e19b4147137ba8f09904730656

(but don't take this patch exactly :laughing: I broke MSRV using `let ... else`)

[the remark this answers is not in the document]
ctz: that looks good to me, will adopt it

ctz:
> One meta-question: what sort of testing have you done for the constant-time properties of the new base64 decoder for the secret data paths? I know this is famously very difficult to nail down concretely when the compiler has so much power to optimize the code differently based on a number of properties, but is there some sanity checking we could do at this stage to gain more confidence? Maybe you already have?

Good question! During development I did look at the disassembly of `CodePoint::decode_secret` to see there were no branches, but that is a bit unsatisfying with a new stable rustc every 6 weeks.

The other thing I tried (just now, in fact) was a test like:

```rust
    #[test]
    fn decode_secret_does_not_branch_or_index_on_secret_input() {
        // this is using the same theory as <https://github.com/agl/ctgrind>
        use crabgrind as cg;

        if matches!(cg::run_mode(), cg::RunMode::Native) {
            std::println!("SKIPPED: must be run under valgrind");
            return;
        }

        let input = b"aGVsbG8gd29ybGQh".to_vec();
        cg::monitor_command(format!(
            "make_memory undefined {:p} {}",
            input.as_ptr(),
            input.len()
        ))
        .unwrap();

        let mut output = [0u8; 32];
        let output = decode_secret(&input, &mut output).unwrap();
        cg::monitor_command(format!(
            "make_memory defined {:p} {}",
            output.as_ptr(),
            output.len()
        ))
        .unwrap();

        std::println!("{output:?}");
    }
```

But this is messy: it confirms there are no branches inside `CodePoint::decode_secret`, but then fires on the `match` that immediately follows it (to skip over whitespace, or error on invalid characters, or count pad characters).

Verifying just `CodePoint::decode_secret` is a bit cleaner:

```rust
    #[test]
    fn codepoint_decode_secret_does_not_branch_or_index_on_secret_input() {
        // this is using the same theory as <https://github.com/agl/ctgrind>
        use crabgrind as cg;

        let input = [b'a'];
        cg::monitor_command(format!(
            "make_memory undefined {:p} {}",
            input.as_ptr(),
            input.len()
        ))
        .unwrap();

        let out = CodePoint::decode_secret(input[0]);
    }
```

And that passes. If you println `out`, then it fails (in the depths of trying to print the secret value), and if you replace `CodePoint::decode_secret` with `CodePoint::decode_public`, you get:

```
==2921557== Thread 2 base64::tests:::
==2921557== Conditional jump or move depends on uninitialised value(s)
==2921557==    at 0x1204ED: rustls_pki_types::base64::CodePoint::decode_public (base64.rs:482)
==2921557==    by 0x13767D: rustls_pki_types::base64::tests::codepoint_decode_secret_does_not_branch_or_index_on_secret_input (base64.rs:748)
```

(As expected, since that function _definitely_ will contain a computed jump or array index on the secret input.)

src/pem.rs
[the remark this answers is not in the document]
ctz: Added as an extra commit on the end, will do some squashing later.

cpu (commented):
Here's some more feedback (almost universally small nits, sorry!) from reviewing this branch to remind myself where we left off.

src/lib.rs
```diff
@@ -757,6 +807,48 @@ impl UnixTime {
...
+    fn from_pem_slice(pem: &[u8]) -> Result<Self, pem::Error> {
+        Self::from_pem_items(&mut pem::read_all_from_slice(pem))
+    }
+
+    /// Decode this type from PEM contained in a `str`.
+    fn from_pem_str(pem: &str) -> Result<Self, pem::Error> {
+        Self::from_pem_slice(pem.as_bytes())
+    }
+
+    /// Decode this type from PEM present in a buffered reader.
```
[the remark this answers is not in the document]
ctz: Reworded doc comment.

src/pem.rs
ctz: Oops, reinstated.

cpu (approved):
Cool!

cpu:
Also noticed that https://github.com/rustls/pemfile/pull/55 is taking this as a dep w/ a git patch but didn't need to change the targetted version.

I think this branch could use a `Cargo.toml` version bump?

src/lib.rs
```diff
@@ -405,6 +482,29 @@ impl TrustAnchor<'_> {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct CertificateRevocationListDer<'a>(Der<'a>);
 
+#[cfg(feature = "alloc")]
+impl DecodePem for Vec<CertificateRevocationListDer<'static>> {
```
djc, line 486: It seems surprising/weird that `DecodePem` is implemented for `PrivatePkcs8KeyDer` but not for `CertificateDer` (and implemented for `Vec<CertificateDer>` but not `Vec<PrivatePkcs8KeyDer>`).

It seems to me that the fundamental required trait method here is more like `from_item(item: pem::Item) -> Option<Self>`.

I think we can rebuild the other things on top of that -- it should compose well with `FromIterator` and things like `Iterator::filter()`.

src/lib.rs
```diff
@@ -920,7 +920,7 @@ impl UnixTime {
 pub trait DecodePem: Sized {
     /// Decode this type from PEM contained in a byte slice.
     fn from_pem_slice(pem: &[u8]) -> Result<Self, pem::Error> {
-        Self::from_pem_items(&mut pem::read_all_from_slice(pem))
+        Self::from_pem_items(&mut pem::Item::iter_from_slice(pem))
```
djc, line 923: Nit: I think we can just import `Item` here?

src/base64.rs
```diff
@@ -0,0 +1,726 @@
...
+        input = new_input;
+
+        let (item, pad) = match f(*byte) {
+            CodePoint::WHITESPACE => continue,
+            CodePoint::INVALID => return Err(Error::InvalidCharacter(*byte)),
+            CodePoint::PAD => (0, 1),
+            CodePoint(n) => (n, 0),
+        };
+
+        if used == 8 {
```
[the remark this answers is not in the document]
ctz: added a comment here.

src/lib.rs
ctz: Yes I agree it currently feels odd, with an implicit assumption that some types are plural (certs, crls) and others singular (keys). Will experiment some more.

djc: Thinking about this more, I feel like `Item` is not quite the right core abstraction -- most people really aren't interested in what it does, which is abstracting over types that can appear in PEM objects. It feels like the core abstraction is that a PEM stream yields `(SectionType, Vec<u8>)` items and then the core pki-types could implement something like this:

```rust
trait FromPem: From<Vec<u8>> {
     const KIND: SectionType;
}
```

ctz: I have been trying this, but it comes a bit unstuck when it comes to `impl FromPem for PrivateKeyDer`, as that has to recognise three possible section types. However I think the previous suggestion still is an improvement -- there's a sketch of this in the commit I just pushed.

djc: Ah yeah, that's a tricky edge case. Maybe the associated const is a slice?

ctz: OK, a little more movement on this. Unfortunately I have abandoned the trait entirely, as it would require rustc 1.75 to be able to return iterators in a trait function. That's a shame. Now there is a macro instead which provides the common constructor/iterator functions.

djc:
@ctz I pushed a commit that replaces the macro with a trait. Have a look and see what you think?

djc (commented):
I think this is a good direction!

cpu (commented):
Keeping an eye on this branch as it evolves. Just had a few small things to surface from my catch-up pass:

fuzz/fuzz_targets/pem.rs
```diff
@@ -4,17 +4,18 @@ use std::io::Cursor;
 
 use libfuzzer_sys::fuzz_target;
 
-use rustls_pki_types::pem;
+use rustls_pki_types::pem::PemObject;
+use rustls_pki_types::{CertificateDer, PrivateKeyDer};
 
 fuzz_target!(|data: &[u8]| {
-    for x in pem::iter_from_reader(&mut Cursor::new(data)) {
+    for x in CertificateDer::pem_reader_iter(&mut Cursor::new(data)) {
```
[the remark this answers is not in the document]
ctz: this is kinda purposeful, to cover the two divergent input paths (std::io, and slices). will add comments here

ctz:
addressed those comments in the latest commit. i'll give some space for any extra comments, and then work on the commit history

djc:
I think I'm satisfied with the direction here, would like to do a last pass after commit history cleanup.

Would maybe suggest trying to rebase the downstream PR in rustls-pemfile before cleaning up commits here.

ctz:
> Would maybe suggest trying to rebase the downstream PR in rustls-pemfile before cleaning up commits here.

That is done, and I've addressed the comments over there.

ctz:
OK, I think the commit history is a bit more sensible.

I did a docs and doctest pass and left that separate for now, but would squash that into the "directly support PEM decoding" commit. But that probably deserves a careful look in case I a word.

cpu (approved):
Tidied up commit history & the updated rustdoc comments all look good to me. I think where this API landed is a really nice end-state. Great work :stars:

src/pem.rs
```diff
@@ -197,36 +197,27 @@ fn read_one_impl(
...
-            let item = match section_type.as_slice() {
-                b"CERTIFICATE" => Some(Item::X509Certificate(der.into())),
-                b"PUBLIC KEY" => Some(Item::SubjectPublicKeyInfo(der.into())),
-                b"RSA PRIVATE KEY" => Some(Item::Pkcs1Key(der.into())),
-                b"PRIVATE KEY" => Some(Item::Pkcs8Key(der.into())),
-                b"EC PRIVATE KEY" => Some(Item::Sec1Key(der.into())),
-                b"X509 CRL" => Some(Item::Crl(der.into())),
-                b"CERTIFICATE REQUEST" => Some(Item::Csr(der.into())),
-                _ => {
+            let section_type = match SectionKind::try_from(&section_type[..]) {
```
djc, line 280: Nit: `section_type` -> `kind`?

src/lib.rs
```diff
@@ -19,7 +19,15 @@
...
 //!
-//! The [rustls-pemfile](https://docs.rs/rustls-pemfile) crate can be used to parse PEM files.
+//! Types here can be created from:
+//!
+//! - DER using (for example) [`PrivatePkcs8KeyDer::from`].
+//! - PEM using (for example) [`PrivatePkcs8KeyDer::from_pem_slice`].
+//!
+//! `from_pem_slice` returns the first matching item from the given input.
+//! `iter_pem_slice` returns every matching item from the given input.
+//! It is usual for a given PEM file to contain multiple items of some types,
```
djc: Nit: suffix function/method calls with `()`?

(I guess most of these are addressed in your docs fixup, but maybe not all of them.)

src/lib.rs
```diff
@@ -123,6 +131,18 @@ impl<'a> PrivateKeyDer<'a> {
     }
 }
 
+#[cfg(feature = "alloc")]
+impl PemObject for PrivateKeyDer<'static> {
+    fn from_pem(kind: SectionKind, value: Vec<u8>) -> Option<Self> {
+        match kind {
```
djc, line 159: So to call this out specifically: should we validate here that the DER is in the correct format? If we don't, I suppose rustls is going to error out at a later stage?

src/pem.rs
```diff
@@ -3,65 +3,178 @@ use alloc::format;
...
+            .unwrap_or(Err(Error::NoItemsFound))
+    }
 
-    /// A DER-encoded plaintext private key; as specified in PKCS #8/RFC 5958
+    /// Iterate over all sections of this type from the PEM contents of the named file.
     ///
-    /// Appears as "PRIVATE KEY" in PEM files.
-    Pkcs8Key(PrivatePkcs8KeyDer<'static>),
+    /// Unlike other `iter_*` functions, this eagerly consumes all the sections
+    /// to bring together errors opening the file and parsing the contents.
```
djc: This is no longer true, right?

src/pem.rs
```diff
@@ -3,65 +3,178 @@ use alloc::format;
...
+    #[cfg(feature = "std")]
+    fn pem_file_iter(
+        file_name: impl AsRef<std::path::Path>,
+    ) -> Result<ReadIter<io::BufReader<std::fs::File>, Self>, Error> {
+        Ok(ReadIter::<_, Self> {
+            rd: io::BufReader::new(std::fs::File::open(file_name).map_err(Error::Io)?),
```
djc, line 55: Nit: we've (conditionally) imported `io`, but are locally qualifying `File`. I think I have a slight preference for also importing `File`.

src/pem.rs
```diff
@@ -3,65 +3,178 @@ use alloc::format;
...
+/// Extract and decode the next supported PEM section from `input`
 ///
 /// - `Ok(None)` is returned if there is no PEM section to read from `input`
 /// - Syntax errors and decoding errors produce a `Err(...)`
-/// - Otherwise each decoded section is returned with a `Ok(Some((Item::..., remainder)))` where
+/// - Otherwise each decoded section is returned with a `Ok(Some((..., remainder)))` where
 ///   `remainder` is the part of the `input` that follows the returned section
-pub fn read_one_from_slice(mut input: &[u8]) -> Result<Option<(Item, &[u8])>, Error> {
+#[allow(clippy::type_complexity)]
+pub fn from_slice(mut input: &[u8]) -> Result<Option<((SectionKind, Vec<u8>), &[u8])>, Error> {
```
djc, line 182: I'm inclined to think that we should keep this private for now? We have a public `SliceIter::new()`.

djc (approved):
So do we now end up with two different `Item` types across this crate and rustls-pemfile? Why is that necessary/helpful? I thought it might be because of inherent methods on `Item`, but it looks like `Item` has no public inherent methods in 2.1.3?

(I also wonder if we should not have `Item` in this crate at all, or maybe make it `#[doc(hidden)]` for now in favor of the newer API.)

src/pem.rs
ctz: I think it would be good to keep this private; I would need to add a `pub fn remainder(&self) -> &[u8] { ... }` (ala the chunk iterators in core) to `SliceIter`

src/pem.rs
ctz: squashed docs fixup commit into this, which fixes this

src/lib.rs
ctz: addressed this remaining one

src/pem.rs
djc: Makes sense. Maybe make that `doc(hidden)`?

src/lib.rs
ctz: Yeah, this is maintaining the status-quo. We could have a `check_format()` or something on `PrivateKeyDer` (extracts the data, guesses the format, and checks that is consistent with the used variant) but it would come short of any sort of key validation or thorough validation of the format -- I think these things need to happen at the level that fully supports DER decoding and cryptography.

djc: > it would come short of any sort of key validation or thorough validation of the format -- I think these things need to happen at the level that fully supports DER decoding and cryptography.

Not sure I agree? The earlier we can signal errors, the more context might be available to the user, especially in this case of format confusion that I imagine would be tricky for inexperienced users. But, it is the status quo so no need to block this PR on it.

PR #54, opened 2024-09-18
Support PEM decoding for `EchConfigListBytes`

ctz:
This adopts the "ECHCONFIG" section delimiter proposed in <https://www.ietf.org/archive/id/draft-farrell-tls-pemesni-05.html>.

(comes after #53)


cpu (approved):
Nice, this looks good.

One thought: I had also started to sketch this out in `rustls/pemfile` a while back. In my [WIP branch](https://github.com/rustls/pemfile/commit/193dde51731b407c0125fb2f20b09a7429b99359) I had also added a `server_ech_configs()` fn that returned `Result<(PrivatePkcs8KeyDer<'static>, EchConfigListBytes<'static>), io::Error>` since I imagine most non-test use-cases will want to load both the private key and the ECH config in one go. 

Do you think there's any value in adding something like that to this branch?

ctz:
> Do you think there's any value in adding something like that to this branch?

I've added a separate commit containing one solution to this; WDYT?

PR #55, opened 2024-09-22
feat: more from impls for `ServerName` and `IpAddr`

nanoqsh:
Let's add more convenient `From` impls.

Now instead of:
```rust
let name = ServerName::IpAddress(IpAddr::V4(Ipv4Addr::LOCALHOST).into());
```
You can write:
```rust
let name = ServerName::from(Ipv4Addr::LOCALHOST);
```

### DIFF
--- a/src/server_name.rs
+++ b/src/server_name.rs
@@ -127,6 +127,45 @@ impl<'a> TryFrom<&'a str> for ServerName<'a> {
     }
 }
 
+impl From<IpAddr> for ServerName<'_> {
+    fn from(addr: IpAddr) -> Self {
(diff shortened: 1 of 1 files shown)


djc (approved):
Makes sense to me, thanks!

djc:
Reminder that while `core::net` now contains `IpAddr` et al, that's a relatively recent development as of 1.77.0, which is why, given our MSRV, we still need to duplicate a bunch of stuff here and guard trait impls with `std` here.

cpu (approved):
Thank you, this is a nice quality of life boost.

PR #60, opened 2024-10-01
Use core::net::Ip{Addr, V4Addr, V6Addr}

akonradi-signal:
Re-export the types from the standard library instead of duplicating their code. The types have been stable since Rust 1.6 and this crate has a MSRV of 1.60.

Instead of re-exporting the equivalent types from core::net at the crate root, this change simply removes the names. This is semver-incompatible, but re-exporting types is also incompatible since a trait implementation in a consuming crate for, e.g., core::net::IpAddr and pki_types::IpAddr would now conflict.


cpu:
> Hm, it was my understanding that it was 1.77 that stabilised core::net?

Mine too, and the [MSRV CI failure](https://github.com/rustls/pki-types/actions/runs/11127874829/job/30921355418?pr=60) hints at similar:

```
error[E0432]: unresolved import `core::net`
 --> src/server_name.rs:8:[11](https://github.com/rustls/pki-types/actions/runs/11127874829/job/30921355418?pr=60#step:4:12)
  |
8 | use core::net::{IpAddr, Ipv4Addr, Ipv6Addr};
  |           ^^^ could not find `net` in `core`
```

> This is semver-incompatible

I think any semver breaking change is a tough sell for this crate. It would be nice to tidy this stuff up but IMO I don't think it merits a breaking release.

djc:
> Hm, it was my understanding that it was 1.77 that stabilised `core::net`?

I think they're talking about when they were standardized in std (as opposed to core).

akonradi-signal:
My mistake, I looked at https://doc.rust-lang.org/stable/core/net/enum.IpAddr.html, which has 1.7. https://doc.rust-lang.org/stable/core/net/index.html says 1.77

PR #65, opened 2024-11-28
fix misc. clippy findings

cpu:
See each commit for more information.


djc (approved):

ctz (approved):
Thanks for doing all these!

PR #67, opened 2024-12-23
Move `AlgorithmIdentifier` values to here from `rustls-webpki`

### LINKED ISSUE #27: Considering adding constant for the PKIX Algorithm Identifier
When implementing a crypto provider only for signature, I'm obligated to add a direct dependency to `webkpi` in order to benefit from the constant of the [webkpi::alg_id] module .  The rustls' verify example for provider does the same [thing](https://github.com/rustls/rustls/blob/d8b918c74dd2a8ad11534c9ffd508dc03662900b/provider-example/src/verify.rs#L7)

It would be nice if `pki-types` do provide this constants because it will ensure we are using the version of `webpki` of whatever the `rustls` crate is using.

Basically, it would mean to migrate the [webkpi::alg_id] module here.
Are you open to that kind of changes?

[webkpi::alg_id]: https://github.com/rustls/webpki/blob/add8fd763024bce80341010cdacf2dec6da46ebf/src/signed_data.rs#L257

ctz:
Any downstream crate defining a custom `rustls::CryptoProvider` currently needs to take a dependency on `rustls-webpki` to reuse these values (alternatively they can redefine these values, but meh.)

Moving them here removes that need, and eliminates another reason for people to depend on `rustls-webpki` directly.

fixes #27


cpu (approved):

djc:
How about we move AlgorithmIdentifier into the new module, with a top-level re-export for compatibility?

djc (approved):

cpu:
@ctz Any thoughts on Djc's last feedback? Should we hold merge on that?

ctz:
> How about we move AlgorithmIdentifier into the new module, with a top-level re-export for compatibility?

Added a commit for this. I could take it or leave it to be honest, but see what you think.

djc (approved):
I kinda like it, though IMO `AlgorithmIdentifier` should go below the `const`s.

PR #68, opened 2025-01-24
feat: add const `try_from_str` constructor to `DnsName` type.

DSharifi:
Add a const `DnsName::from_str` method to create `DnsName` in a const setting. 

`validate()` is refactored to use while loop instead of for loop as for loops are non const.


cpu (approved):
Thanks :+1: Seems reasonable to me!

Would you mind writing a commit message? It could be as simple as the PR description.

DSharifi:
Thanks for the quick review!

I Amended the commit message now to the PR title.

src/server_name.rs
```diff
@@ -204,6 +204,14 @@ impl<'a> DnsName<'a> {
             Err(_) => Err(s),
         }
     }
+
+    /// Produces a borrowed [`DnsName`] from a borrowed [`str`].
+    pub const fn try_from_str(s: &str) -> Result<DnsName<'_>, InvalidDnsNameError> {
+        match validate(s.as_bytes()) {
+            Ok(_) => Ok(DnsName(DnsNameInner::Borrowed(s))),
+            Err(_) => Err(InvalidDnsNameError),
```
djc, line 212: Suggest yielding the error returned by `validate()` rather than reinstantiating the same value.

Can we replace the implementation in `TryFrom<&str>` with a call to this function to avoid duplication?

src/server_name.rs
```diff
@@ -323,7 +331,9 @@ fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
         return Err(InvalidDnsNameError);
     }
 
-    for ch in input {
+    let mut character_index = 0;
```
djc, line 334: I would prefer a terser variable name, maybe `idx`?

Seems a little unfortunate to use the indexing which is potentially more prone to panicking... I wonder whether the compiler can elide the bounds checks here.

DSharifi: > I would prefer a terser variable name, maybe idx?

Renamed it to `idx`. However, I do prefer `character_index` + `character` over `idx` + `ch` as it makes the code much clearer to read.

DSharifi: > Seems a little unfortunate to use the indexing which is potentially more prone to panicking... I wonder whether the compiler can elide the bounds checks here.

Agreed that it is unfortunate that we have to index directly. The only other alternative I can think of of that would work is using recursion, but I am not sure it's worth the complexity it adds:
```rust
const fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
    const MAX_LABEL_LENGTH: usize = 63;
    const MAX_NAME_LENGTH: usize = 253;

    if input.len() > MAX_NAME_LENGTH {
        return Err(InvalidDnsNameError);
    }

    enum State {
        Start,
        Next,
        NumericOnly { len: usize },
        NextAfterNumericOnly,
        Subsequent { len: usize },
        Hyphen { len: usize },
    }

    const fn validate_recursive(input: &[u8], state: State) -> Result<(), InvalidDnsNameError> {
        use State::*;
        match input {
            [] => match state {
                Start | Hyphen { .. } | NumericOnly { .. } | NextAfterNumericOnly => {
                    Err(InvalidDnsNameError)
                }
                _ => Ok(()),
            },
            [ch, rest @ ..] => {
                let next_state = match (state, ch) {
                    (Start | Next | NextAfterNumericOnly | Hyphen { .. }, b'.') => {
                        return Err(InvalidDnsNameError)
                    }
                    (Subsequent { .. }, b'.') => Next,
                    (NumericOnly { .. }, b'.') => NextAfterNumericOnly,
                    (Subsequent { len } | NumericOnly { len } | Hyphen { len }, _)
                        if len >= MAX_LABEL_LENGTH =>
                    {
                        return Err(InvalidDnsNameError)
                    }
                    (Start | Next | NextAfterNumericOnly, b'0'..=b'9') => NumericOnly { len: 1 },
                    (NumericOnly { len }, b'0'..=b'9') => NumericOnly { len: len + 1 },
                    (Start | Next | NextAfterNumericOnly, b'a'..=b'z' | b'A'..=b'Z' | b'_') => {
                        Subsequent { len: 1 }
                    }
                    (Subsequent { len } | NumericOnly { len } | Hyphen { len }, b'-') => {
                        Hyphen { len: len + 1 }
                    }
                    (
                        Subsequent { len } | NumericOnly { len } | Hyphen { len },
                        b'a'..=b'z' | b'A'..=b'Z' | b'_' | b'0'..=b'9',
                    ) => Subsequent { len: len + 1 },
                    _ => return Err(InvalidDnsNameError),
                };
                validate_recursive(rest, next_state)
            }
        }
    }

    validate_recursive(input, State::Start)
}
```

djc: Recursion definitely doesn't seem like an improvement.

PR #69, opened 2025-01-24
feat: add const try_from_str constructor to DnsName type.

DSharifi:
reopening https://github.com/rustls/pki-types/pull/68 as it got auto closed.

PR #70, opened 2025-01-27
Bump version to 1.11.0

DSharifi:
### What's Changed
- A `const` method, `DnsName::from_str`, was added by @DSharifi in https://github.com/rustls/pki-types/pull/69.
* Move `AlgorithmIdentifier` values to here from `rustls-webpki` by @ctz in https://github.com/rustls/pki-types/pull/67


djc:
Please rebase instead of merging.

I think the release notes should be more like this:

* Migrate `AlgorithmIdentifier` here from rustls-webpki, to allow third-party `CryptoProvider`s to avoid taking a dependency on rustls-webpki.
* Add a `const` `ServerName::try_from_str()` method to allow callers to build `ServerName` values in `const` contexts.

cpu (approved):

djc:
So I guess we can just close this?

PR #73, opened 2025-02-08
allow domains of length 254 with a trailing dot

### LINKED ISSUE #72: `DnsName::try_from_str` incorrectly disallows domains of length 254 that have a trailing `.`
As the below code shows, `DnsName::try_from_str` disallows domains of length 254 that have a trailing `.`:

```rust
use rustls_pki_types::DnsName;
fn main() {
    // This proves that `DnsName::try_from_str` allows and even retains trailing `.`.
    assert!(DnsName::try_from_str("example.com.").map_or(false, |dom| dom.as_ref() == "example.com."));
    let long_label =
        "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
    assert_eq!(long_label.len(), 63);
    // The maximum length of a domain is 254 if a trailing `'.'` exists; otherwise the max length is 253.
    let mut long_domain = format!("{long_label}.{long_label}.{long_label}.{long_label}");
    long_domain.pop();
    long_domain.pop();
    long_domain.push('.');
    assert_eq!(long_domain.len(), 254);
    // This incorrectly `panic`s despite being a valid domain.
    assert!(DnsName::try_from_str(long_domain.as_str()).is_ok());
}
```

`DnsName::try_from_str` allows for trailing dots; therefore the maximum length of a `DnsName` is 254, not 253, when there is a trailing dot. The maximum length is 253 only when there isn't a trailing dot. This can easily be explained by the fact that the maximum length of a domain _in wire format_ is 255 since domains in wire format are serialized such that each label is preceded by one byte that represents its length. This is necessary since DNS is an octet protocol that allows _all_ octets to exist in a label; furthermore, the only label that is allowed a length of 0 is the root label. This root label is necessary so a resolver knows when there are no more labels in a domain.

When domains are in a _representation format_—which is what `DnsName::try_from_str` is based on—that disallows "escape characters", then it's one of two forms:

1. `<label_1>.<label_2>...<label_n>`
2. `<label_1>.<label_2>...<label_n>.`

When _both_ 1. and 2. are transformed into wire format, they produce the same output:

`[label_1_len, label_1, label_2_len, label_2, ..., label_n_len, label_n, 0]` where _label_i_len_ is the one byte value inclusively between 1 and 63 that represents the number of bytes that make up _label_i_ and _label_i_ is shorthand for each byte that it consists of. From this, we see that the length of such a domain is the sum of lengths of each label plus the total number of labels plus 1.

Now we need to derive a function that maps the length of 1. and 2. to the length of the wire-format value since it is the wire format value that matters. We are assuming that the representation format is "simple" (e.g., there are no "escape characters" and each "character" is encoded in 1 byte (e.g., ASCII)).

1. The total length of the domain is equal to the sum of the lengths of each label plus _n - 1_. This is _2_ less than the _actual_ domain in wire format. Since the max length of a domain in wire format is _255_, this means the max length is _255 - 2 = 253_.
2. The total length of the domain is equal to the sum of the lengths of each label plus _n_. This is _1_ less than the _actual_ domain in wire format. Since the max length of a domain in wire format is _255_, this means the max length is _255 - 1 = 254_.

If the above is all too technical and "math-y", then here are "proofs by example":

Using [`domain`](https://docs.rs/domain/0.10.3/domain/index.html) which is a library for DNS libraries (i.e., it conforms to [RFC 9499](https://www.rfc-editor.org/rfc/rfc9499.html) and related RFCs). It is maintained by the same company that maintains popular DNS software like [Unbound](https://www.nlnetlabs.nl/projects/unbound/about/) and [NSD](https://www.nlnetlabs.nl/projects/nsd/about/): 

```rust
use domain::base::Name;
fn main() {
    let long_label = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
    assert_eq!(long_label.len(), 63);
    let mut long_domain = format!("{long_label}.{long_label}.{long_label}.{long_label}");
    long_domain.pop();
    long_domain.pop();
    long_domain.push('.');
    assert_eq!(long_domain.len(), 254);
    assert!(Name::<Vec<u8>>::from_chars(long_domain.chars()).is_ok());
}
```

Using [`idna`](https://docs.rs/idna/latest/idna/index.html) which is a library for [IDNA](https://url.spec.whatwg.org/#idna) according to WHATWG URL Standard:

```rust
use idna::uts46::{AsciiDenyList, DnsLength, Hyphens, Uts46};
fn main() {
    let long_label = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
    assert_eq!(long_label.len(), 63);
    let mut long_domain = format!("{long_label}.{long_label}.{long_label}.{long_label}");
    long_domain.pop();
    long_domain.pop();
    long_domain.push('.');
    assert_eq!(long_domain.len(), 254);
    assert!(Uts46::new()
        .to_ascii(
            long_domain.as_bytes(),
            // This is a proper superset of characters that are allowed by [`DnsName::try_from_str`].
            // This is not relevant though since we are simply showing that domains of length 254 with
            // a trailing `.` are valid.
            AsciiDenyList::URL,
            // Even though `DnsName::try_from_str` disallows hyphens in certain positions, we don't here.
            // Again, this is irrelevant for this example since we are dealing with a domain that only contains
            // the letter a.
            Hyphens::Allow,
            DnsLength::VerifyAllowRootDot
        )
        .is_ok());
}
```

Using [`ascii_domain`](https://docs.rs/ascii_domain/0.6.2/ascii_domain/) which is a library maintained by me that deals strictly with "representation format" domains:

```rust
use ascii_domain::{char_set::PRINTABLE_ASCII, dom::Domain};
fn main() {
    let long_label = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
    assert_eq!(long_label.len(), 63);
    let mut long_domain = format!("{long_label}.{long_label}.{long_label}.{long_label}");
    long_domain.pop();
    long_domain.pop();
    long_domain.push('.');
    assert_eq!(long_domain.len(), 254);
    // `PRINTABLE_ASCII` is a proper superset of what `DnsName::try_from_str` allows, but that is
    // irrelevant for this example since we are dealing with a domain that only contains
    // the letter a.
    assert!(Domain::try_from_bytes(long_domain.as_bytes(), &PRINTABLE_ASCII).is_ok());
}
```

zacknewman:
Fixes #72.

### DIFF
--- a/src/server_name.rs
+++ b/src/server_name.rs
@@ -847,7 +847,7 @@ mod tests {
         ("numeric-only-middle-label.4.com", true),
         ("1000-sans.badssl.com", true),
         ("twohundredandfiftythreecharacters.twohundredandfiftythreecharacters.twohundredandfiftythreecharacters.twohundredandfiftythreecharacters.twohundredandfiftythreecharacters.twohundredandfiftythreecharacters.twohundredandfiftythreecharacters.twohundredandfi", true),
-        ("twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourc.", true),
+        ("twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfour.", true),
         ("twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourcharacters.twohundredandfiftyfourc", false),
     ];
 
--- a/src/server_name.rs
+++ b/src/server_name.rs
@@ -331,11 +331,15 @@ const fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
 
     match input.len() {
         0 | 255.. => return Err(InvalidDnsNameError),
-        // This is a `const fn`, so we cannot use
-        // `Option::unwrap_or_else(|| unreachable!("bug in [u8]::len()")).
-        MAX_NAME_LENGTH_WITH_TRAILING_DOT if *input.last().unwrap() != b'.' => {
-            return Err(InvalidDnsNameError)
-        }
+        MAX_NAME_LENGTH_WITH_TRAILING_DOT => if let Some(lst) = input.last() {
+            if *lst != b'.' {
+                return Err(InvalidDnsNameError);
+            }
+        } else {
+            // Ideally this would be `unreachable`, but that can't happen until
+            // MSRV is updated to a version where it is `const`.
+            panic!("there is a bug in [u8]::len()");
+        },
         _ => (),
     }
 
--- a/src/server_name.rs
+++ b/src/server_name.rs
@@ -320,27 +320,43 @@ const fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
     use State::*;
     let mut state = Start;
 
-    /// "Labels must be 63 characters or less."
+    /// The maximum length of a label in wire format is 63 octets per
+    /// [RFC 1035 § 2.3.4](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.4).
     const MAX_LABEL_LENGTH: usize = 63;
 
-    /// The maximum lenght of a domain name is 253 when there is no trailing dot.
-    /// When there is a trailing dot, the maximum length is 254. This is true since
-    /// a domain with a trailing dot gets transformed into the same wire-format value
-    /// as one without.
-    const MAX_NAME_LENGTH_WITH_TRAILING_DOT: usize = 254;
+    /// The maximum length of a domain in wire format is 255 octets per
+    /// [RFC 1035 § 2.3.4](https://www.rfc-editor.org/rfc/rfc1035#section-2.3.4).
+    const MAX_NAME_LENGTH: usize = 255;
 
-    match input.len() {
-        0 | 255.. => return Err(InvalidDnsNameError),
-        MAX_NAME_LENGTH_WITH_TRAILING_DOT => if let Some(lst) = input.last() {
-            if *lst != b'.' {
-                return Err(InvalidDnsNameError);
+    match input.last() {
+        // Empty domains are not valid.
+        None => return Err(InvalidDnsNameError),
+        // `input` is a domain in _representation format_, but length requirments are based on the
+        // wire format. For representation formats that forbid the root domain (represented as either `b" "` or
+        // `b"."`), allow a trailing `b'.'`, and where each "character"/`u8` in a label is transformed to a `u8`
+        // (e.g., there are no "escape characters"); then the formula is simple: add `1` when there is a trailing
+        // `b'.'` otherwise add `2`.
+        //
+        // Examples
+        //
+        // * "ab.c" gets transformed to the wire format [0x02, 0x61, 0x62, 0x01, 0x63, 0x00]. As we see the length
+        //   of the representation format is 4 and the length of the wire format is 6 and 4 + 2 = 6.
+        // * "ab.c." gets transformed to the wire format [0x02, 0x61, 0x62, 0x01, 0x63, 0x00]. As we see the length
+        //   of the representation format is 5 and the length of the wire format is 6 and 5 + 1 = 6.
+        //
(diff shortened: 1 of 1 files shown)


src/server_name.rs
```diff
@@ -323,11 +323,24 @@ const fn validate(input: &[u8]) -> Result<(), InvalidDnsNameError> {
     /// "Labels must be 63 characters or less."
     const MAX_LABEL_LENGTH: usize = 63;
 
-    /// https://devblogs.microsoft.com/oldnewthing/20120412-00/?p=7873
-    const MAX_NAME_LENGTH: usize = 253;
-
-    if input.len() > MAX_NAME_LENGTH {
-        return Err(InvalidDnsNameError);
+    /// The maximum lenght of a domain name is 253 when there is no trailing dot.
```
djc, line 326: Do you have chapter and verse from the relevant RFCs on this?

djc: So what problem are you actually solving here? It all seems pretty theoretical to me.

ctz: > This means you literally cannot use `rustls` as a client-side TLS software suite if you ever wanted to access domains that it does not deem valid but in fact are.

Do you have a practical example of such a name -- specifically one that is allowed by RFC4985, issued in a certificate under cabforum BR 7.1.2.7.12, etc?

> `DnsName::try_from_str` does not allow `" "` which is a valid domain per that section.

To be clear the `DnsName` type we have here exists to support the `ServerName` type, and `ServerName` contains the name or address for a TLS server from the view of a client. Therefore, there is little to be gained to allow a client to express not-useful values like `" "`, even if they are technically allowed and useful in DNS software.

I'm definitely open to updating references here to (probably) refer to https://www.rfc-editor.org/rfc/rfc9499.html#name-names -- a BCP which reflects modern usage better than RFC1035

cpu: https://docs.rs/rustls-pki-types/latest/rustls_pki_types/enum.IpAddr.html

https://github.com/rustls/rustls/blob/fa3e31746d5721b60df4d407c5df346d4ec7e423/ROADMAP.md?plain=1#L100-L106

briansmith:
> For example, if Firefox ever wanted to use rustls for its TLS-handling, it literally couldn't without sacrificing the ability to access certain domains that it can now (e.g., Firefox happily accesses domains that contain ; in a label).

Hi, I wrote both Firefox's certificate DNS name parsing code and also (IIUC) this pki-types code. You can see the current Firefox source code with the same "253" limit here: https://searchfox.org/mozilla-central/source/security/nss/lib/mozpkix/lib/pkixnames.cpp#1916

Also, you are discussing the IP address stuff with the people who implemented IP address support in this library and/or for Let's Encrypt, just FYI.

briansmith:
> I would like to construct a DnsName that is a valid domain whose labels consist of any printable ASCII (sans space and .) or perhaps at a minimum what WHATWG says is allowed 

It is intentional that the WHATWG how-to-deal-with-URLs-in-HTML specification is not used to decide how to parse domain names in certificates in either Firefox or any of these Rust libraries. I believe it is true that there are certain domain names that Firefox and other browsers will try to connect to, but they've become pretty uniform regarding rejecting certificates with names that differ from what's enforced here. So, there may be some domain names that you could connect to using http:// but not https:// because certificate validation would always fail. And, also, the CA would immediately get in trouble for issuing certificates with those domains, if it is a public CA.

briansmith:
> If I am unable to get Chromium or Firefox 

I would say that Chromium probably is setting the upper bound on how lax they would be for non-public CA certificates, as Chrome and Edge actually have "enterprise" customers that they care about. You can look at the source code in Chromium to see how they configure their certificate verifier (which I think is in BoringSSL now). I doubt there is any appetite to get any laxer than that.

djc:
Section 3 of [RFC 6066](https://www.rfc-editor.org/rfc/rfc6066#section-3) includes this language:

> "HostName" contains the fully qualified DNS hostname of the server, as understood by the client.  The hostname is represented as a byte string using ASCII encoding without a trailing dot.

As such, we could instead add a check that the given string does not include a trailing dot, but increasing the allowed length to accomodate a trailing dot is clearly incorrect for the stated purpose of representing a RFC 6066 `HostName` (which is what this type is intended for).

briansmith:
When the user types "https://example.org." in the address bar, the client should give the certificate verification library the string "example.org." as the domain name. The certificate verification library should match a dNSName "example.org" in a certificate 
with "example.org." as though it were "example.org". That is, for the purposes of certificate verification, the trailing dot in the reference identifier is ignored, even though it isn't ignored for cookie handling and other things.

The certificate verifier should reject a certificate with a dNSName of "example.org." as trailing dots in dNSName are not allowed, IIRC.

ctz:
> As such, we could instead add a check that the given string does not include a trailing dot, but increasing the allowed length to accomodate a trailing dot is clearly incorrect for the stated purpose of representing a RFC 6066 `HostName` (which is what this type is intended for).

Note that rustls already strips a trailing period if needed to meet this clause of RFC6066.

djc:
Closing this for lack of progress.

PR #74, opened 2025-02-21
rustfmt: style_edition 2024

ctz:
(also did `cargo fix --edition` and `cargo fix --edition-idioms`, with no changes resulting)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -327,7 +327,7 @@ impl PrivatePkcs1KeyDer<'_> {
 #[cfg(feature = "alloc")]
 impl zeroize::Zeroize for PrivatePkcs1KeyDer<'static> {
     fn zeroize(&mut self) {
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -327,7 +327,7 @@ impl PrivatePkcs1KeyDer<'_> {
 #[cfg(feature = "alloc")]
 impl zeroize::Zeroize for PrivatePkcs1KeyDer<'static> {
     fn zeroize(&mut self) {
-        self.0 .0.zeroize()
```
cpu, line 330: I'm happy to see this get fixed. The old way is so ugly!

cpu (approved):
Thanks!

src/lib.rs
ctz: me too! totally grot

PR #76, opened 2025-03-27
Add cargo deny check in CI

djc:
As discussed in https://github.com/rustls/rustls-ffi/pull/554#issuecomment-2752525908. This both checks the lockfile against the RustSec advisory database and checks dependency licenses as well as some other stuff. I've used this in many other projects (like Quinn), and Brian also set it up for webpki.


cpu (approved):
Thanks!

Nit-picking slightly, it might be nicer to separate the `--locked` commit & the `Cargo.lock` addition from the `deny.toml` & `audit` CI addition.

ctz (approved):

PR #78, opened 2025-04-13
Add AlgorithmIdentifiers for ML-DSA variants

djc:
Not sure whether `ML_DSA` or `MLDSA` is the more idiomatic way to spell these. IMO if the prose spelling includes a hyphen, then a word separator in the Rust identifiers is to be expected, but [aws-lc-rs currently spells this MLDSA_44](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/unstable/signature/static.MLDSA_44.html).

From https://www.ietf.org/archive/id/draft-ietf-lamps-dilithium-certificates-07.html#name-identifiers. It's a little unclear what that means for the permanence of these identifiers? I get the impression they're being quoted here (also, stuff like https://oid-base.com/get/2.16.840.1.101.3.4.3.17 exists) but it's not completely clear to me what the normative reference is supposed to be.


ctz:
> From https://www.ietf.org/archive/id/draft-ietf-lamps-dilithium-certificates-07.html#name-identifiers. It's a little unclear what that means for the permanence of these identifiers? I get the impression they're being quoted here (also, stuff like https://oid-base.com/get/2.16.840.1.101.3.4.3.17 exists) but it's not completely clear to me what the normative reference is supposed to be.

I think the primary source is https://csrc.nist.gov/projects/computer-security-objects-register/algorithm-registration#DSA but given the US government, anything from the IETF is preferable (even a random draft of an ID).

ctz (approved):
I think "ML-DSA" is the canonical spelling.

I regenerated these from first principles with graviola, and they match.

PR #79, opened 2025-05-01
Zeroize on drop

djc:
#71 added `Zeroize` impls for a bunch of types. It did not add zeroization on drop by default, which we do use for a bunch of stuff in rustls itself. It did expose the `Zeroize` trait as part of the public API, unlike rustls, which is probably not a good idea for this crate. Add a `cargo check-external-types` run to CI -- although it seems like it would not have caught this issue.


src/lib.rs
```diff
@@ -283,10 +274,20 @@ impl TryFrom<Vec<u8>> for PrivateKeyDer<'_> {
...
 
     fn try_from(key: Vec<u8>) -> Result<Self, Self::Error> {
-        Ok(match PrivateKeyDer::try_from(&key[..])? {
-            PrivateKeyDer::Pkcs1(_) => Self::Pkcs1(key.into()),
-            PrivateKeyDer::Sec1(_) => Self::Sec1(key.into()),
-            PrivateKeyDer::Pkcs8(_) => Self::Pkcs8(key.into()),
+        let key_ref = PrivateKeyDer::try_from(&key[..])?;
+        Ok(match &key_ref {
+            PrivateKeyDer::Pkcs1(_) => {
+                drop(key_ref);
```
djc, line 280: This slightly awkward restructuring is to avoid bothering the borrow checker, which otherwise complains that the borrowing `PrivateKeyDer` value needs to run `Drop` at the end of the function, after moving `key` into the owning `PrivateKeyDer`.

ctz: Is this sort of thing likely to happen also in calling code? (I can't really find any clear declaration that impling Drop is a minor change, but I do have a bit set that it varies the borrow checker behaviour in some circumstances -- [this perhaps?](https://doc.rust-lang.org/std/ops/trait.Drop.html#drop-check) -- but it isn't mentioned at all in the semver rules.)

djc: I think it could happen, but would probably be unlikely? I'm having a hard time coming with up with a specific example.

Given the lack of specific mention of `impl Drop` in the semver guidelines, I guess it should be treated like any other trait impl, and thus be classified as minor? I filed [an issue](https://github.com/rust-lang/cargo/issues/15471) against the SemVer reference guide.

dare3path:
You probably know this already by now, but if left as is(b50af82), the PR due to `Drop`, would cause `rcgen` to fail to compile as:

```rust
/home/user/SOURCE/github/rcgen
$ cargo test
    Updating crates.io index
     Locking 1 package to latest compatible version
    Updating rustls-pki-types v1.11.0 (/home/user/SOURCE/github/modified_forks/pki-types) -> v1.12.0
   Compiling rustls-pki-types v1.12.0 (/home/user/SOURCE/github/modified_forks/pki-types)
   Compiling x509-parser v0.17.0
   Compiling botan-sys v0.11.1
   Compiling rcgen v0.14.0 (/home/user/SOURCE/github/rcgen/rcgen)
   Compiling rustls-webpki v0.103.1
error[E0597]: `private_key` does not live long enough
   --> rcgen/src/key_pair.rs:200:31
    |
199 |         let private_key = pem::parse(pem_str)._err()?;
    |             ----------- binding `private_key` declared here
200 |         let private_key_der: &[_] = private_key.contents();
    |                                     ^^^^^^^^^^^ borrowed value does not live long enough
201 |         Self::from_pkcs8_der_and_sign_algo(&PrivatePkcs8KeyDer::from(private_key_der), alg)
    |                                             ----------------------------------------- a temporary with access to the borrow is created here ...
202 |     }
    |     -
    |     |
    |     `private_key` dropped here while still borrowed
    |     ... and the borrow might be used here, when that temporary is dropped and runs the `Drop` code for type `PrivatePkcs8KeyDer`
    |
    = note: the temporary is part of an expression at the end of a block;
            consider forcing this temporary to be dropped sooner, before the block's local variables are dropped
help: for example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block
    |
201 |         let x = Self::from_pkcs8_der_and_sign_algo(&PrivatePkcs8KeyDer::from(private_key_der), alg); x
    |         +++++++                                                                                    +++

error[E0597]: `private_key` does not live long enough
   --> rcgen/src/key_pair.rs:291:27
    |
290 |         let private_key = pem::parse(pem_str)._err()?;
    |             ----------- binding `private_key` declared here
291 |         let private_key: &[_] = private_key.contents();
    |                                 ^^^^^^^^^^^ borrowed value does not live long enough
292 |         Self::from_der_and_sign_algo(
293 |             &PrivateKeyDer::try_from(private_key).map_err(|_| Error::CouldNotParseKeyPair)?,
    |              ------------------------------------------------------------------------------ a temporary with access to the borrow is created here ...
...
296 |     }
    |     -
    |     |
    |     `private_key` dropped here while still borrowed
    |     ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `ControlFlow<Result<Infallible, error::Error>, PrivateKeyDer<'_>>`
    |
    = note: the temporary is part of an expression at the end of a block;
            consider forcing this temporary to be dropped sooner, before the block's local variables are dropped
help: for example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block
    |
292 ~         let x = Self::from_der_and_sign_algo(
293 |             &PrivateKeyDer::try_from(private_key).map_err(|_| Error::CouldNotParseKeyPair)?,
294 |             alg,
295 ~         ); x
    |

For more information about this error, try `rustc --explain E0597`.
error: could not compile `rcgen` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
```

djc:
Yeah -- so apparently this is a major change, so I guess we can't just add it like this. And I think it's also probably not worth doing a 2.0 only for this?

ctz:
Yeah I'm afraid this alone doesn't motivate a 2.0 for me :(

For now we could keep the Zeroize impls, and downstream users can write `Zeroizing<PrivateKeyDer<'_>>`. That is a bit unsatisfying though.

ctz:
I think the latter is the best way to go -- specifically provide some zeroizing newtypes  (maybe generic) types that hold a `PrivateKey*` value. That means downstream users need to opt-in to the drop behaviour change.

ctz (approved):

cpu (approved):

djc:
* Published rustls-pki-types v1.12.0 at registry `crates-io`
* [new tag]         v/1.12.0 -> v/1.12.0
* [Release notes](https://github.com/rustls/pki-types/releases/tag/v%2F1.12.0)

PR #81, opened 2025-06-19
Mark all possible functions const

ctz:
Did this at the start of march and then forgot about it.

My ultimate goal was to be able to to write `const SOMETHING: UnixTime = UnixTime::since_unix_epoch(...)` which is not a strong motivation. But AIUI this is free, and mostly mechanical.

(Can't enable `clippy::missing_const_for_fn` which points out these changes, as it doesn't respect MSRV.)

### DIFF
--- a/src/base64.rs
+++ b/src/base64.rs
@@ -203,7 +203,7 @@ impl CodePoint {
         )
     }
 
-    fn decode_public(a: u8) -> Self {
+    const fn decode_public(a: u8) -> Self {
         const TABLE: [CodePoint; 256] = [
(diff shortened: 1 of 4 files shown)


djc:
> (Can't enable clippy::missing_const_for_fn which points out these changes, as it doesn't respect MSRV.)

Should file an issue about this...

djc (approved):

ctz:
> Should file an issue about this...

-> https://github.com/rust-lang/rust-clippy/issues/15079

PR #82, opened 2025-07-01
Add `From<DnsName<'a>>` for `ServerName<'a>`

stormshield-gt:
When using `DnsName` with `rustls`, they are several places where I need the conversion to a `ServerName`.

This PR adds a `From<DnsName<'a>>` for `ServerName<'a>` which IMHO is nicer to use than `ServerName::DnsName(..)` because we can just do `into()`


djc (approved):
Makes sense. Can you add a commit fixing the clippy issues? Just `--fix` should do the job.

PR #96, opened 2025-12-16
Add algorithm id for ECDSA with secp256k1 curve 

mkmks:
At @zama-ai, we're building an FHE coprocessor for confidential smart contracts on Ethereum. This coprocessor relies on an MPC network that protects FHE private keys, which we also built. Our MPC nodes have Ethereum identities, and we found it useful to use the exact same identity keys to establish the nodes' mTLS identities to not make the MPC network dependent on centralized PKI.

Using ECDSA with the secp256k1 curve that Ethereum uses isn't commonly supported by TLS libraries, such as `rustls`, so we had to make sure `rustls` supports it. That wasn't hard because `rustls` already relies on `aws_lc_rs`, which supports ECDSA with secp256k1.

To make it work, we only had to add  `ECDSA_NISTP256K1_SHA256` to the list of the supported signature schemes in `rustls`. But that required adding `ECDSA_NISTP256K1_SHA256` as a supported signature scheme in the `rustls-webpki` crate, which, in turn required adding the `ECDSA_P256K1` algorithm identifier in the `rustls-pki-types` crate.

This PR does exactly that and nothing more: it creates the `ECDSA_P256K1` identifier. It's the first of the three PRs that would upstream this whole endeavour. On its successful merge, the related PRs against `rustls-webpki` and `rustls` could follow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -91,7 +91,7 @@ dependencies = [
 
 [[package]]
 name = "rustls-pki-types"
-version = "1.13.1"
+version = "1.13.2"
 dependencies = [
  "crabgrind",
  "web-time",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustls-pki-types"
-version = "1.13.1"
+version = "1.13.2"
 edition = "2021"
 rust-version = "1.60"
 license = "MIT OR Apache-2.0"


djc:
> Also true that `rustls` has no public API that would allow for using arbitrary algorithms from the underlying cryptographic libraries, such as `aws_lc_rs`. If that were the case, we wouldn't have to propose adding the definitions to support our use case.

@ctz's earlier wording was confusing, but all the necessary APIs are in fact public. You can create your own rustls `CryptoProvider` which references your own `&'static dyn SignatureVerificationAlgorithm`.

djc:
> Thanks for the suggestion, I'll look into it, and perhaps, it wouldn't be necessary to patch `webpki` after all.

If it is, we'd consider that a bug to solve!

djc:
[Released](https://github.com/rustls/pki-types/releases/tag/v%2F1.13.2).

PR #103, opened 2026-02-03
Remove mention of rustls-pemfile from docs

ranile:
This is not present in the README either and the link in lib.rs leads to a crate belonging to a repository that has already been archived.

Linking to `rustls-pemfile` after https://github.com/rustls/pemfile/issues/61 feels incorrect


djc (approved):

djc:
Thanks!

PR #110, opened 2026-06-30
1.15.0: ensure `FipsStatus::Certified` is usable 

ctz:
(another option here is to remove the `#[non_exhaustive]` from the variant. This choice allows for items for the variant data to be added in the future, with the restriction that they must have some sensible value for default that can be used by the constructor fn.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -91,7 +91,7 @@ dependencies = [
 
 [[package]]
 name = "rustls-pki-types"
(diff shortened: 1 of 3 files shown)


cpu (approved):
> another option here is to remove the #[non_exhaustive] from the variant

I think I lightly prefer adding the constructor like you have here.

djc (approved):

PR #112, opened 2026-07-22
Faster DNS name validation

djc:
While looking at benchmark results for

- https://github.com/rustls/rustls/pull/3163

I noticed that `rustls_pki_types::server_name::validate()` was more expensive than I expected. Worked with Claude to optimize it (and verify that it has the same results). Criterion results from this improvement:

```
     Running benches/dns_name.rs (target/release/deps/dns_name-bef6433e6c7ace85)
Gnuplot not found, using plotters backend
dns_name/typical        time:   [124.84 ns 125.63 ns 126.45 ns]
                        change: [−55.558% −55.144% −54.712%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
dns_name/ip_like_rejected
                        time:   [20.163 ns 20.318 ns 20.484 ns]
                        change: [−27.528% −26.440% −25.354%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
dns_name/max_length     time:   [132.86 ns 137.70 ns 143.23 ns]
                        change: [−68.711% −67.456% −66.337%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,7 +391,7 @@ checksum = "d6f6ff9a378485b298a5286656da665ba74413d36db0979633275d2e708145d4"
 
 [[package]]
 name = "rustls-pki-types"
-version = "1.15.0"
+version = "1.15.1"
 dependencies = [
  "crabgrind",
  "criterion",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,52 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "1.1.4"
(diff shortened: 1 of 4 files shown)


ctz (commented):
Can we have both the performance the formal state-machine type? What happens if we have the state machine type without the large fields?

ctz (approved):

